PR #2, opened 2017-10-18
Check license

Koubek:
- The license must be placed inside the script root filesystem.
 - Solution for Microsoft/nav-docker#75


freddydk:
Thanks.
This was implemented slightly differently.
Now the license file is copied to a shared folder if needed.

PR #13, opened 2017-11-07
added Start-NavContainerAppDataUpgrade

waldo1001:


freddydk:
Thanks.
Nice one - I was actually going to create this one today, but never got around to it.

waldo1001:
Well - it's not perfect yet to work with these.
I think we should include the -version and -Publisher and other parameters to be able to identify a specific published app for synching, installing and upgrading.. .

At this point, we need to uninstall, unpublish and publish every time we want to upgrade an app.. .  
Which might be fine for Docker-stuff (test and dev) ... but is definitely not fine for production .. .

PR #15, opened 2017-11-07
Ability to create CSide environment but not have to export objects

MikeGlue:
Use new -exportObjectsToText switch.


freddydk:
A few things

1. please turn around the switch: -dontExportObjects (as this makes all the object handling functions stop working)

2. Please do not have a big IF statement and two new-navcontainer calls in new-csidedevcontainer. Instead use -dontExportObjects:$dontExportObjects in one call.

3. if -dontExportObjects is set - avoid the license check in new-navcontainer

MikeGlue:
Feedback implemented. Thanks.

freddydk:
Thanks - I will include this in the next update

Issue #14, opened 2017-11-07
sql timeout error sqlcmd


sodgeronimo:
first of all, my issue is resolved by a workaround (been searching since sunday on what the problem was, so i thought id share)

scenario:
laptop with windows 10 pro, latest stable docker version and navcontainerhelper 0.1.1.4.

i tried to make a new container with the following powershell script 

```
New-CSideDevContainer `
    -containerName 'DEVServerPreviewFINUS' `
    -devImageName 'microsoft/dynamics-nav:devpreview-finus' `
    -licenseFile 'D:\Workfolder\Licenties\License.flf' `
    -vmAdminUsername admin `
    -adminPassword (convertTo-SecureString -String $password -AsPlainText -Force) `
    -memoryLimit 3G `
    -updateHosts `
    -auth NavUserPassword `
    -accept_eula `
    -Verbose 
```
This gives me the following output: 
```

Creating Nav container DEVServerPreviewFINUS
Using image microsoft/dynamics-nav:devpreview-finus
Using license file D:\Workfolder\Licenties\License.flf
NAV Version: 11.0.18712.0-finus
Generic Tag: 0.0.3.2
Creating container DEVServerPreviewFINUS from image microsoft/dynamics-nav:devpreview-finus
Waiting for container DEVServerPreviewFINUS to be ready, this shouldn't take more than a few minutes
Time:          ½              1              ½              2
..............................docker : Sqlcmd: Error: Microsoft ODBC Driver 13 for SQL Server : Named Pipes Provider: Could not open a connection to SQL Server [2]. .
At C:\Program Files\WindowsPowerShell\Modules\navcontainerhelper\0.1.1.4\NavContainerHelper.psm1:992 char:21
+             $logs = docker logs $containerName
+                     ~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Sqlcmd: Error: ...L Server [2]. .:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

```
The container gets created and is accessible when you open it manually.
There are however no shortcuts created.

The problem is with the export of the objects.

the problem lies in the name of the container, from what i can see this can be a max of 15 characters because when you open the cside client it shows the following: DEVSERVERPREVIE\SQLEXPRESS

the remaining characters of the host name are no longer present.

I removed and recreated the container a few times and ultimately tried again with DEVPrevFINUS as a container name and this didnt give me the error anymore.

[...]

PR #18, opened 2017-11-09
SQLCMD fails on named pipes, changed to TCP

irijcken:
SQLCMD consistently fails to connect on my W10 laptop when trying to use named pipes. Even when using Enter-NavContainer and trying to run the same SQLCMD from the PS prompt inside the container it does not succeed to connect. When I force SQLCMD to connect through TCP the script runs and continues successfully. 

When the SQL connection fails, many things after this command also fail (creation of desktop shortcuts, export of objects, etc...). The container does reach healthy status however and everything else seems to working OK.

sqlcmd : Sqlcmd: Error: Microsoft ODBC Driver 13 for SQL Server : Named Pipes Provider: Could not open a connection to SQL Server [2]. .
    + CategoryInfo          : NotSpecified: (Sqlcmd: Error: ...L Server [2]. .:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Sqlcmd: Error: Microsoft ODBC Driver 13 for SQL Server : Login timeout expired.
Sqlcmd: Error: Microsoft ODBC Driver 13 for SQL Server : A network-related or instance-specific error has occurred while establishing a connection to SQL Server. Server is not found or not accessible. Check if instance name is correct and if SQL Server is config
ured to allow remote connections. For more information see SQL Server Books Online..

### DIFF
--- a/NavContainerHelper.psm1
+++ b/NavContainerHelper.psm1
@@ -836,7 +836,7 @@ function New-NavContainer {
         $programFilesFolder = Join-Path $containerFolder "Program Files"
         New-Item -Path $programFilesFolder -ItemType Directory -ErrorAction Ignore | Out-Null
 
-        'sqlcmd -d $DatabaseName -Q "update [dbo].[Object] SET [Modified] = 0"
+        'sqlcmd -S tcp:127.0.0.1\sqlexpress -d $DatabaseName -Q "update [dbo].[Object] SET [Modified] = 0"
         ' | Add-Content -Path "$myfolder\AdditionalSetup.ps1"
         
         if (Test-Path $programFilesFolder) {


freddydk:
Your pull request crossed a refactoring checkin from me, where all functions was split out in seperate files.
Could you redo your change and I will have a look.

freddydk:
BTW - this wouldn't by any chance be related to https://github.com/Microsoft/navcontainerhelper/issues/14
would it?

freddydk:
I see that in other places i use
sqlcmd -S 'localhost\SQLEXPRESS' ....
I guess that would be the same - wouldn't it?

sodgeronimo:
won't that be a problem when you specify a databaseserver through the additional parameters?
I havent checked yet in the code on how this is handled, just a use case i've been thinking on

freddydk:
That is correct @sodgeronimo - in fact that SQLCMD should not run unless you are using localhost\SQLEXPRESS.

freddydk:
I have fixed this in navcontainerhelper 0.1.2.0 - Ivo - let me know if this doesn't solve your problem

irijcken:
Problem solved, noticed that my PR crossed your check-in but I was unable to respond earlier due to other responsibilities. Good job!

PR #30, opened 2017-11-28
Update New-CSideDevContainer.ps1

dNsl9r:
Fix example


freddydk:
Thanks.
I never use new-csidedevcontainer myself - i use new-navcontainer with -includecside (does the same thing).
New-CsideDevContainer is kept for compatibility reasons (and it is mentioned in a lot of places) - I will mark it as deprecated soon.

PR #52, opened 2017-12-11
Upgrade-specific functions from waldo's module.

waldo1001:
There are functions that I heavily used in terms of upgrading a NAV database (CU or Major upgrade).  I think they might come in use in combination with docker as well - although they probably need some enhancements.
I hope this will be useful for you.


freddydk:
Thanks a million Waldo - I will merge those into a development branch and hope to be able to use a lot from this.

DanielGoehler:
@freddydk We would like to utilize [Copy-NAVEnvironment.ps1](https://github.com/microsoft/navcontainerhelper/pull/52/commits/f698aff24dc2c6cff243dd18bc41663f2f74c51f#diff-f9e38f682889fc1bec5dcdbaeaf9381604ba441ba2c965ed6966bd1a6c043b24). Any change you could look into this PR?

freddydk:
Spinning up another service tier inside a docker container would not be the right thing to do - what exactly are you trying to do?
The PR was originally created by Waldo to allow me to take learnings and some of these functions and incorporate them in ContainerHelper.

DanielGoehler:
Yes, you are right. I didn't notice that this happens inside of the container. I still try to find the best way to copy a container. SQL Backup and Restore seems the best way to do it.

tfenster:
Depending on your scenario, another (faster) approach could be to just stop the first container, copy the database files and start the second container with some logic to attach the database files

freddydk:
Yes, like Tobias said - or use an external SQL server (maybe on the host) for the database - then shutting down one BC container and spinning up the next.

PR #62, opened 2017-12-12
Automatic service startup

MikeGlue:
When using local SQL server within the container, set the required services to automatic startup in the event the container is stopped/started.

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -247,6 +247,16 @@ function New-NavContainer {
         ') | Add-Content -Path "$myfolder\AdditionalSetup.ps1"
     }
 
+    # Set services to automatic start when using local container server
+    'if ($databaseServer -eq "localhost" -and $databaseInstance -eq "SQLEXPRESS") {
+        Set-Service ''W3SVC'' -startuptype Automatic
+        Set-Service ''MSSQL$SQLEXPRESS'' -startuptype Automatic
+        Set-Service ''SQLTELEMETRY$SQLEXPRESS'' -startuptype Automatic
+        Set-Service ''SQLWriter'' -startuptype Automatic
+        Set-Service ''SQLBrowser'' -startuptype Automatic
+        Set-Service ''MicrosoftDynamicsNavServer$NAV'' -StartupType Automatic
+    }' | Add-Content -Path "$myfolder\AdditionalSetup.ps1"
+
     Write-Host "Creating container $containerName from image $imageName"
 
     if ($useSSL) {


freddydk:
Not sure I understand why you would do this?
Upon restart, the scripts are run, and services will restart?
The reason for doing this scriptet is, that the container might have a new hostname (you might have committed the container and cloned it) and then the service tier would have to have a new SSL certificate and be restarted during initializatoin.
Also, a scriptet start ensures that we know exactly when services are started - and you can specify a new database to a committed container.

Did you encounter any problem in the current behavior or is this purely a performance enhancement?

MikeGlue:
The reason I had to do this was because the SQL and NAV services were set to manual startup mode, and when I stopped and started the container those services never came back on. I'm using this on a laptop so I need to stop and start containers frequently as resources are limited. This change makes the assumption that if you are using the SQL Express within the container, then you would also want the services to always run after a container restart.

freddydk:
It is correct that SQL, NAV and IIS are set to manual start, but stopping and starting the container should re-run start.ps1 and thus the services are started by the script.
If you are experiencing that this doesn't happen, I would like to find out why?
Docker logs on my machine will look like this after a stop and a start:

<end of docker run output>
Files:
http://test:8080/al-0.12.15355.vsix

Initialization took 74 seconds
Ready for connections!
Initializing...
Restarting Container
PublicDnsName unchanged
Hostname is test
PublicDnsName is test
Using NavUserPassword Authentication
Starting Local SQL Server
Starting Internet Information Server
Starting NAV Service Tier
Container IP Address: 172.19.158.188
Container Hostname  : test
Container Dns Name  : test
Web Client          : http://test/NAV/WebClient/
Dev. Server         : http://test
Dev. ServerInstance : NAV

Files:
http://test:8080/al-0.12.15355.vsix

Initialization took 30 seconds
Ready for connections!

Initialization is run a second time and services are restarted (if necessary)

MikeGlue:
Hmm strange, I created a brand new container tonight and I now get the expected results the same as what you get. Using the container I created late last week using the exact same script, I do not get the same results. Seems to skip right over the service restarting.

Initialization took 60 seconds
Ready for connections!
Initializing...
Restarting Container
PublicDnsName unchanged
Hostname is mgDev365
PublicDnsName is mgDev365
Using NavUserPassword Authentication
Container IP Address: 172.24.36.34
Container Hostname  : mgDev365
Container Dns Name  : mgDev365
Web Client          : https://mgDev365/NAV/WebClient/
Dev. Server         : https://mgDev365
Dev. ServerInstance : NAV

I'm closing this pull request in any case and will keep an eye on things to see if this issue crops up again.

freddydk:
Could you send me a docker inspect of the "old" container?

MikeGlue:
[DockerInspect.txt](https://github.com/Microsoft/navcontainerhelper/files/1568836/DockerInspect.txt)

I have deleted my container that had the issue but confirmed on a colleague's machine that the issue existed in one of their containers. The docker inspect for that container is attached.  I'm always pulling the newest NavContainerHelper functions so something must've been fixed since the version published late last week.

PR #96, opened 2018-03-06
Use settings file for RTC client desktop shortcut

marknitek:
To prevent the RTC client from using the shared **clientusersettings.config** file (in _%appdata%\Microsoft\Microsoft Dynamics NAV\110_), 
we should provide the argument **-settings:ClientUserSettings.config**.
This way the shortcut always works even when working with multiple containers on the same machine.


freddydk:
Thanks Marco

PR #99, opened 2018-03-14
Added function Get-NavContainerServerConfiguration.

richardrobberse:
Added function Get-NavContainerServerConfiguration, this function will retrieve the server configuration of the server in the container and return it as an PowerShell object.

The idea was based on the Get-NAVServerConfiguration2 from Waldo but was tweaked so we can access the settings like this: Configuration.DatabaseInstance, Configuration.DatabaseName etc.


freddydk:
I like the idea.
A few things:
- help is missing for parameters.
- no need to have a serverinstance parameter - it will always be NAV inside a container
- pipelining isn't used anywhere else, I would like to add that to all in one go - and not start off with one function and then break something if I decide to do things differently
- code doesn't look like the other functions in this module (begin, end, resultobjectarray). Now that might be because we should change all the other functions, but I would like to keep things the same and then refactor in one go.

freddydk:
I have taken the function from this PR and added to the next version of navcontainerhelper

PR #102, opened 2018-03-20
Create a table of content with links

tfenster:
I see myself using this a lot as reference for my own scripts, so having a TOC instead of doing search and find would help a lot


freddydk:
Cool - just learned something new...

PR #124, opened 2018-04-17
Get-NAVDockerImageName

waldo1001:
Just a small addition, based on your blog: https://blogs.msdn.microsoft.com/freddyk/2018/04/16/which-docker-image-is-the-right-for-you/

If it's not useful enough, I'll add it in my NCH extension ;-)

### DIFF
--- a/Misc/Get-NAVDockerImageName.ps1
+++ b/Misc/Get-NAVDockerImageName.ps1
@@ -1,0 +1,78 @@
+function Get-NAVDockerImageName {
+    <#
+    .SYNOPSIS
+    Compiles a Docker Image string
+
+    .EXAMPLE
+    Get-NAVDockerImageName -ImageType CurrentNAV -Cu 3 
+
+    Returns: microsoft/dynamics-nav:cu3
+
+    .EXAMPLE
+    Get-NAVDockerImageName -ImageType CurrentNAV -VersionOrBuild 2018 -Country BE -Cu 4
+
+    Returns: microsoft/dynamics-nav:2018-cu4-be
+
+    .EXAMPLE
+    Get-NAVDockerImageName -ImageType CurrentBusinessCentral -VersionOrBuild "12.0.21229.0" -Country be    
+    
+    Returns: microsoft/bcsandbox:12.0.21229.0-be
+
+    .EXAMPLE
+    Get-NAVDockerImageName -ImageType NextBusinessCentral -Country DK
+    
+    Returns: bcinsider.azurecr.io/bcsandbox:dk
+
+    .EXAMPLE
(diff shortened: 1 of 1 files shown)


freddydk:
A few comments
cu's do not always start with cu (rtm is a good example of that, devpreview-october was another)
else the function will work for now.
I am still debating with myself whether I think the function is a good idea.

waldo1001:
Another idea is may be to get to a list to all combinations on a specific repository?

waldo1001:
Quickly put together a PowerShell script to scrape all tags from the DockerHub repo....

`
$ResultingObject = @()

$result = Invoke-WebRequest -Uri "https://registry.hub.docker.com/v2/repositories/microsoft/dynamics-nav/tags/" 
$JsonObject = ConvertFrom-Json -InputObject $result.Content
$ResultingObject += $JsonObject.results

while ($JsonObject.next) {
    $result = Invoke-WebRequest -Uri $JsonObject.next 
    $JsonObject = ConvertFrom-Json -InputObject $result.Content
    $ResultingObject += $JsonObject.results
    
    Write-host "Count: $($ResultingObject.Count)"
}

$ResultingObject.name
`

richardrobberse:
@waldo1001 I've created a couple of functions to achieve the same thing - you can find them [here](https://github.com/richardrobberse/blog/blob/master/NavDockerAPIHelper.psm1)
Get-NavDockerImage will retrieve all the images.
Test-NavDockerImage will check if a specified image is valid.

I asked Freddy if they're a useful addition to the navcontainerhelper, but I guess they are :)

waldo1001:
@richardrobberse - Like it! :-)  If @freddydk doesn't add it, I'm always willing to add it to my repo ;-).

freddydk:
The problem here is really that I don't think people should be querying for which containers exist.
If you locate 12.0.21500.0 and 12.0.21700.0 - then you will automatically assume that #2 is newer than #1.
Well - they might be from different branches and the newest might be 11.0.21439.0 - which tomorrow might be 13.0.21732.0 (because we updated the version number).
I think people should use the images described on the blog post and only if specifically instructed to do so, they should go for a specific build.
That is the reason why I am hesitant to put this in. I will be the one who has to answer questions on what is newest and why 11 or 12 or 13 (has already happened)

waldo1001:
Well, if version numbers doesn't really mean too much - it makes sense to only very specifically refer to it.. .  If listing them would cause confusing - let's not list them.
I was merely trying to demystify... but this doesn't seem demystifyable on the level of navcontainerhelper... ;-)

freddydk:
Did you already add this to your modules?

waldo1001:
I have, but in the meantime forgot about it, and never updated it - and never use it.
I will update it.. .

PR #126, opened 2018-04-18
Added possibility to choose which tenant database to export when exporting bacpacs

richardrobberse:
By default the database 'tenant' is being exported - this is the template tenant.
Any changes done in the container (Default database) are never included in the exported tenant bacpac.

In our case we want to maintain our setup/sample data in a Docker container and at some point export this data and use it for new deployments.

Note: this only applies to multi-tenant containers.


freddydk:
I have renamed the parameter and removed the validateset and will take the change like this.
You could easily have people creating tenants for different versions of demo data,

Issue #113, opened 2018-03-29
Use a local image


peterz84:
Is it possible to use a local image (manually pulled from navinsider.azurecr.io) ? Or to use an image from this server with username and password login?

Using this commands:

docker login "navinsider.azurecr.io" -u "xyz" -p "abc"
$PSScriptRoot = "D:\Documents\Docker\MyFirstApp"
$mylicense = "D:\license.flf"
$imageName = "navinsider.azurecr.io/dynamics-nav:11.0.21063.0-finnl"
$sourceFolder = Join-Path $PSScriptRoot "Source"
$containerName = Split-Path $PSScriptRoot -Leaf
New-NavContainer -accept_eula  `
                 -alwaysPull -imageName $imageName `
                 -containerName $containerName `
                 -memoryLimit 3G -auth Windows `
                 -licensefile $mylicense `
                 -updateHosts `
                 -includeCSide `
                 -additionalParameters @("--volume ${sourceFolder}:c:\source") 

And getting this error:
docker.exe: invalid reference format: repository name must be lowercase.

freddydk:
The above is correct and should work as written.
you can try new-navcontainer -accept_eula -imagename $imagename -containername test
with the above value for imagename - that should just work.
On the error it sounds like you have had a capital letter in the imagename.

peterz84:
It seems something is wrong with -additionalParameters @("--volume ${sourceFolder}:c:\source")
Omitting this line helps. But the question is why.

freddydk:
Do you have a subfolder called Source in the folder from which you are running the script?

peterz84:
I do. I think the problem was the path was too long. I moved everything to a location with a shorter path and now it works.

PR #134, opened 2018-05-18
Parameter enableSymbolLoading added for Issue #113

megel:
I have added the Parameter enableSymbolLoading :)


freddydk:
New-CSideDevContainer is likely to be removed - new-navcontainer should be used instead

PR #172, opened 2018-07-26
Added support for "Version" in sync, publish, install & uninstall Nav…

hemisphera:
Installing multiple versions of the same app does not seem to be supported. So now "Publish-NavContainerApp" reads the version from the app and passes it along to Sync / Install. Synch / Install / and UnInstall-NavContainerApp now support an optional parameter "appVersion".

### DIFF
--- a/AppHandling/Install-NavContainerApp.ps1
+++ b/AppHandling/Install-NavContainerApp.ps1
@@ -26,7 +26,16 @@ function Install-NavContainerApp {
     $session = Get-NavContainerSession -containerName $containerName
     Invoke-Command -Session $session -ScriptBlock { Param($appName, $appVersion, $tenant)
         Write-Host "Installing $appName on $tenant"
-        Install-NavApp -ServerInstance NAV -Name $appName -Version $appVersion -Tenant $tenant
+        $parameters = @{ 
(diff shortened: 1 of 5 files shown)


freddydk:
This seems to be a breaking change.
Everybody with scripts will have to add appVersion as it is madatory.
The parameter will need to be made optional - and only transferred to the underlying functions if specified.
Thanks

hemisphera:
I don't think this is a breaking change. The new parameters I have introduced are **not** mandatory.

You are correct though, that they are always transferred to the underlying function. But if they are not specified, they will be transferred as `$null` and the underlying functions already handle that case correctly.

hemisphera:
I have just tested this and it works. The app "test" I am synchronizing below doesn't even exist. So the cmdlet seems to be pretty forgiving ...

image

I just wanted to avoid duplicating code, but if you prefer, I will change it to an if-else block.

freddydk:
I prefer a construct like this:

```
$params = @{ }
if ($version) {
  $params += @{ "version" = $version }
}
sync-navapp nav -name $name @params
```

/Freddy

freddydk:
Thanks, will include this.

PR #179, opened 2018-08-07
Remove symbol generation; move sync after objects import. Resolves #163

### LINKED ISSUE #163: Importing testtoolkit ends with timeout on schema synchronization
On some images I am getting this error when creating container including test toolkit>
```
: [31719426] The objects from the .fob file were imported, but they contain schema changes that cannot be synchronized to the database with the Microsoft Dynamics NAV Server instance.
Contact your system administrator with the following information:

Server Name:         localhost
Server Instance:     NAV
Tenant Name:

Description:
The requested task failed to start within the timeout period of 10 seconds that has been specified in the Microsoft Dynamics NAV Development Environment / Database / Alter. The
reason is that Microsoft Dynamics NAV Server is busy with a similar task.
 -- File: C:\TestToolKit\CALTestRunner.fob
```
I cannot see any way how to change the timeout. Right now I am getting this on image microsoft/dynamics-nav:2018-ch. It is mostly on images with heavy localization (more objects?).

MikeGlue:
Remove symbol generation; move sync until after all objects are imported. Resolves #163

### DIFF
--- a/ObjectHandling/Import-TestToolkitToNavContainer.ps1
+++ b/ObjectHandling/Import-TestToolkitToNavContainer.ps1
@@ -13,7 +13,7 @@
 #>
 function Import-TestToolkitToNavContainer {
     Param(
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [string]$containerName, 
         [System.Management.Automation.PSCredential]$sqlCredential = $null
     )
@@ -29,7 +29,6 @@ function Import-TestToolkitToNavContainer {
         $databaseInstance = $customConfig.SelectSingleNode("//appSettings/add[@key='DatabaseInstance']").Value
         $databaseName = $customConfig.SelectSingleNode("//appSettings/add[@key='DatabaseName']").Value
         if ($databaseInstance) { $databaseServer += "\$databaseInstance" }
-        $enableSymbolLoadingKey = $customConfig.SelectSingleNode("//appSettings/add[@key='EnableSymbolLoadingAtServerStartup']")
     
         $params = @{}
         if ($sqlCredential) {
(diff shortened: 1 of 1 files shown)


freddydk:
How come you don't need the symbol generation?
How can you use the symbols in the objects?

MikeGlue:
The 'Test' symbol package is already part of the out of box Cronus database so there is no need to regenerate them.

freddydk:
Really???
I thought I tried functions like assert, and they where not there, but I might be wrong

Få Outlook til iOS<https://aka.ms/o0ukef>

Fra: Mike Glue <notifications@github.com>
Sendt: tirsdag, august 7, 2018 8:32 PM
Til: Microsoft/navcontainerhelper
Cc: Freddy Kristiansen; Comment
Emne: Re: [Microsoft/navcontainerhelper] Remove symbol generation; move sync after objects import. Resolves #163 (#179)


The 'Test' symbol package is already part of the out of box Cronus database so there is no need to regenerate them.

MikeGlue:
As I just did a pull request for a function to generate the symbols, I'll close this and update it to make use of that function so user can choose to generate symbols or not. In my case, I'm importing various objects during my build process and I don't want to generate symbols until they've all been imported.

freddydk:
Generating symbols on import only generates for the imported objects as I understand. A full generate is much slower.

Få Outlook til iOS<https://aka.ms/o0ukef>

Fra: Mike Glue <notifications@github.com>
Sendt: tirsdag, august 7, 2018 9:42 PM
Til: Microsoft/navcontainerhelper
Cc: Freddy Kristiansen; Comment
Emne: Re: [Microsoft/navcontainerhelper] Remove symbol generation; move sync after objects import. Resolves #163 (#179)


As I just did a pull request for a function to generate the symbols, I'll close this and update it to make use of that function so user can choose to generate symbols or not. In my case, I'm importing various objects during my build process and I don't want to generate symbols until they've all been imported.

MikeGlue:
I'm not sure, but if I use the 'generateSymbolReference' option when using the C/Side client, when I import objects it almost always locks up the service tier. I'm not sure if it is doing the symbol generation 1 object at a time and it just is too much, but when I remove the generate symbol option and just do a normal import without it, all is fine, and generating the symbols after that is only a few minutes.

PR #180, opened 2018-08-07
Symbol generation command

MikeGlue:
Command to generate application symbols in a NAV container.


freddydk:
Will take this - but I will rename the function to Generate-xxx instead of Create-xxx

Issue #182, opened 2018-08-14
Test Symbols will be not downloaded in Compile-AppInNavContainer


ChrisBlankDe:
If a version for test is defined in app.json the specific symbols version must be downloaded.
The Download Link could be build like the system and Applikation package:
https://server:7049/nav/dev/packages?publisher=Microsoft&appName=Test&versionText=11.0.0.0

https://github.com/Microsoft/navcontainerhelper/blob/e876571f1369afd66b5728e9bea38d9c884e457d/AppHandling/Compile-AppInNavContainer.ps1#L72

freddydk:
I have accepted the PR - will be included in the next update (within a week).

PR #183, opened 2018-08-14
Add test symbols as dependency

ChrisBlankDe:
This should solve #182 
Test Symbols will be added as dependency if "test" node ~~exists~~ is not empty in app.json


freddydk:
Thanks for this PR

PR #191, opened 2018-08-30
Added support for "Mode" in Sync-NavContainerApp

hemisphera:
I've added the paramter "Mode" that gets passed along to the underlying Sync-NAVApp cmdlet. Default is "Add", just as stated in MS documentation.


freddydk:
I prefer Parameter(Mandatory=$false) to make it very clear.
And  $mode should have its own if statement - meaning that if it isn't provided, it is not added to the parameters.

hemisphera:
You are correct, of course. I didn't think that a parameter with `ValidateSet` would be nullable. I've fixed that.

PR #220, opened 2018-10-10
Added support for "TenantSpecificProperties" in Get-NavContainerAppInfo

hemisphera:
Just as the title says.

Btw: I don't know if you prefer the notation **t**enantSpecificProperties  over **T**enantSpecificProperties. I usually use uppercase, but you mostly use camelCase, so I changed that.


freddydk:
I have included this in my dev branch, will be checked in today

PR #226, opened 2018-10-16
Added support for StyleCop and improved Convert-ALCOutputToAzureDevOps

IceOnly:
Compile-AppInNavContainer can now run the StyleCop check with the parameter EnableCodeCop.
With the new parameter FailOn, it is now possible to let fail the build pipeline in AzureDevOps (based on warnings and/or errors occurred during the compile).
- none (default) - same as before, only output of the errors and warnings
- warning - fail build if a warning or error occurred
- error - fail build if a error occurred


freddydk:
Thanks

PR #280, opened 2018-12-07
Add parameters scope and syncMode

ajkauffmann:
New parameters in Publish-NavContainerApp:

- scope - to enable publish and install as tenant extension, which is similar to how AL extension in VS Code works.
- syncMode - to enable Add, Clean or Development mode for database schema synchronization. The default sync mode used by VS Code is Development, to enforce database synchronization without increasing the version number of the app.

### DIFF
--- a/AppHandling/Publish-NavContainerApp.ps1
+++ b/AppHandling/Publish-NavContainerApp.ps1
@@ -12,12 +12,16 @@
   Include this parameter if the app you want to publish is not signed
  .Parameter sync
   Include this parameter if you want to synchronize the app after publishing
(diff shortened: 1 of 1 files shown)


freddydk:
Looks good, thanks

kine:
After this merge, there is error when using the code>

ajkauffmann:
My fault.... I was quickly redoing the changes after a clean clone of the repo. Saw Sync-..... and put the Mode in the wrong place.

freddydk:
I have checked in a fix for this to make the new parameters optional instead of defaulted.

PR #286, opened 2018-12-12
Add syncTenantMode Parameter

eknraw:
-Mode Parameter for Sync-NavTenant


freddydk:
Is this needed at all?

This implementation suggestion becomes a breaking change, which we should avoid.

These things needs to be implemented so that leaving the parameter out will also leave the parameter out downstream - or default the parameter to the default value downstream.

PR #291, opened 2018-12-18
improve access check

tfenster:
improve the access check by not only checking for elevation but also access to the Docker Engine named pipe if the current session is not elevated

Incredibly ugly but I couldn't find a better way

### DIFF
--- a/NavContainerHelper.psm1
+++ b/NavContainerHelper.psm1
@@ -1,4 +1,4 @@
-﻿Set-StrictMode -Version 2.0
+Set-StrictMode -Version 2.0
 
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks for the investigations and suggestion - I will build in this check instead:
$dockerOk = $true
try {
    $ps = (docker ps)
    if ($LASTEXITCODE -ne 0) {
        $dockerOk = $false
    }
} catch{
    $dockerOk = $false
}
if (!$dockerOk) {
    Write-Warning "NavContainerHelper might not work. Current user might not have permissions or docker engine might not be running."
}

PR #294, opened 2018-12-20
fixed format for servers with different culture

ChrisBlankDe:
On none English servers the function generates "Time"-Values like "260,336" which will be interpreted by DevOps as 3 days instead of 4 minutes.

The added parameter will ignore the local format and always export with "." (dot) as decimal marker.


freddydk:
The function is a temporary implementation until we have a better way of getting test results - but I agree with this PR.

PR #296, opened 2018-12-27
Suggest to add portainer script 

lhtan-dev:
Adding portainer creation script. Does not include validation check on existing containers with portainers name created.


freddydk:
Sorry, but I will decline this. I do not want to extend NavContainerHelper to support Portainer. I think it is great that people use portainer or the BCinaB (https://github.com/tfenster/BCinaB/wiki/Installation-and-Usage) project from Tobias, but I am not going to add dependencies in NavContainerHelper to these projects.

PR #330, opened 2019-01-30
DevOps error detection ALC internal error

matejskubic:
Added case to report error to DevOps for ALC internal errors like:
`error AL0999: Internal error: System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException...`


freddydk:
@kine - my knowledge about regex is limited, could you give me your review on this PR?

kine:
@freddydk The regex will match anything having "error XXNNNN: " text inside where XX is two times a-z, A-Z, 0-9 or _ and NNNN is 4 times number. It should cover the example and I think it is not in conflict with some possible text which could be detected as false positive (nothing come to my mind).

freddydk:
Thanks

PR #336, opened 2019-02-04
Copy-CompanyInNavContainer feature

dkatson:
Hi @freddydk ,
This feature is useful during testing process, when I want to copy company from Cronus f.ex. to new company, and then test some business process there.


freddydk:
Could you rename the function to Copy-NavCompanyInNavContainer - and make the parameters the same names as the Copy-NavCompany function - then I will take it (SourceCompanyName and DestinationCompanyName) - I try to do this will all the "proxyfunctions"

freddydk:
Thanks

PR #337, opened 2019-02-04
EscapeDataString for $publisher to handle "&"

StefanMaron:
if the app publisher is the company and the comapany name is like "GmbH & Co. KG" the web request fails


freddydk:
Thanks

Issue #325, opened 2019-01-29
Compile-AppInNavContainer symbol files for Microsoft_Application are missing customization 


ichladil:
I am using container with enabled symbol loading and imported Test Libraries. 
The app.json does not include Test.
Compilation via Compile-AppInNavContainer is failing on the missing test libraries references.

I have checked it a bit further:
When manually downloading symbols via VS Code the Microsoft_Application includes the test libraries.
When using Compile-AppInNavContainer the downloaded Microsoft_Application is missing the test libraries.
Version of the Microsoft_Application also differs in those scenarios. 

I assume this is caused by versionText parameter in Invoke-RestMethod which in VS Code is taken from app.json - as it is (13.0.0.0). 
Whereas in Compile-AppInNavContainer the versionText for Application seems to be overriden by Version provided by Get-NavAppInfo -SymbolsOnly section (13.0.26556.0 in my case of CU2)

freddydk:
The Test symbols are special I think - doesn't VS Code get these into a specific file?

ichladil:
Test symbols shall not be handled in the special way if the objects are imported into BC using generatesymbolreferences (and that is the case in my example). 
My main point is that Application symbols exported from BC container when using Compile-AppInNavContainer exports out-of-the-box DVD symbol references only ignoring the fact that I have modified via C/SIDE or via objects import some further objects into BC application.

If the line 
`$version = $app.version`
is executed only if publisher <> 'Microsoft' or name <> Application the result would be perhaps correct even in this scenario.

freddydk:
I am having a hard time reproing this - I might be doing something wrong???

NAVFreak:
This could be divided into two issues:
1. Compile-AppInNavContainer doesn't compile with the same symbols as VS Code does (my pull request)
2. The imported CSIDE objects are not included in the symbol files when choosing the same app version as the same one installed. 

For number 2 i have filed a support request to Microsoft.

freddydk:
I just talked to the guys from the modern dev team.
The interesting thing is to find out how you got the two different versions of the symbols in there in the first place.
If you run Get-NavContainerAppInfo with -symbolsOnly what do you get?
I get: 
```
ServerInstance : MicrosoftDynamicsNavServer$NAV
PSComputerName : a863168e31d2a77ade77af5a4ca3cd8bf9298de61eed88202a736cfef4a8da64
[...]

PR #353, opened 2019-02-19
GetSymbolsUsingApp.json

NAVFreak:
This deals with issues:
[https://github.com/Microsoft/nav-docker/issues/325#issuecomment-460238675](https://github.com/Microsoft/nav-docker/issues/325#issuecomment-460238675)
[https://github.com/Microsoft/navcontainerhelper/issues/325](https://github.com/Microsoft/navcontainerhelper/issues/325)

The major problem is that you get different symbols when downloading in VS code  and using Compile-AppInNavContainer, it should be the same.

The problem is appearing when you have these settings in app.json and are using a newer docker image:
```
  "platform": "13.0.0.0",
  "application": "13.0.0.0",
```

If you use image 'mcr.microsoft.com/businesscentral/onprem:se' and download sysmbols you se the following in the terminal:
```
[2019-02-19 08:40:15.97] Sending request to http://dockermall:7049/NAV/dev/packages?publisher=Microsoft&appName=Application&versionText=13.0.0.0
[2019-02-19 08:40:15.99] Sending request to http://dockermall:7049/NAV/dev/packages?publisher=Microsoft&appName=System&versionText=13.0.0.0
```  
And you get the files:
_Microsoft_Application_13.0.27183.0.app
Microsoft_System_13.0.12929.0.app_

Now delete the symbol files in .alpackages folder, copy your app folder to your docker container:
_C:\ProgramData\NavContainerHelper\Extensions\$ContainerName\my_

```
$Containername = '<ContainerName>'
$UserName = '<UserName>'
$PassWord = '<Password>'
$Credential = (New-Object System.Management.Automation.PSCredential($UserName,(ConvertTo-SecureString $PassWord -AsPlainText -Force)))
Compile-AppInNavContainer -containerName $ContainerName `
    -appProjectFolder "C:\ProgramData\NavContainerHelper\Extensions\$Containername\my\App" `
    -appOutputFolder "C:\ProgramData\NavContainerHelper\Extensions\$Containername\my" `
    -UpdateSymbols `
    -credential $Credential

```
Now go to the app folder och look the symbols in the folder:
_C:\ProgramData\NavContainerHelper\Extensions\DockerMall\my\App\.alpackages_
You get the files:
_Microsoft_Application_13.3.27233.0.app
Microsoft_System_13.0.12929.0.app_

### DIFF
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -30,22 +30,22 @@
 #>
 function Compile-AppInNavContainer {
     Param(
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [string]$containerName,
-        [Parameter(Mandatory=$false)]
+        [Parameter(Mandatory = $false)]
         [string]$tenant = "default",
-        [Parameter(Mandatory=$false)]
+        [Parameter(Mandatory = $false)]
         [System.Management.Automation.PSCredential]$credential,
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [string]$appProjectFolder,
-        [Parameter(Mandatory=$false)]
+        [Parameter(Mandatory = $false)]
         [string]$appOutputFolder = (Join-Path $appProjectFolder "output"),
-        [Parameter(Mandatory=$false)]
+        [Parameter(Mandatory = $false)]
         [string]$appSymbolsFolder = (Join-Path $appProjectFolder ".alpackages"),
         [switch]$UpdateSymbols,
         [switch]$AzureDevOps,
         [switch]$EnableCodeCop,
-        [ValidateSet('none','error','warning')]
+        [ValidateSet('none', 'error', 'warning')]
         [string]$FailOn = 'none'
     )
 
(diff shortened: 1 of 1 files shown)


freddydk:
I think the change is correct - but a few comments.

Reformatting source code should not be done as part of a bugfix - makes it harder to see what has changed and what not.

There is something wrong in the fix.
$Version is undefined - meaning that it will of course be -ne $appVersion and -ne $platformversion.
and the code will never hit the "old" line. I think the right fix is to always to with the info in app.json.

NAVFreak:
Ohh, I didn't think about the reformatting :(  
Ran it just to make sure my code was properly pasted in.

Isn't $version assigned on line 113..118?
```
$app = $publishedApps | Where-Object { $_.publisher -eq $publisher -and $_.name -eq $name }
if ($app) {
    $version = $app.version
}
```
There is one thing missing though and that is the correct name of the symbol file. As we have seen we ask for symbol to a certain version (13.0.0.0) but can ge gived another version. This fix doesn't handle that. A proper fix would have been to open up the symbol package and extract it from the NavxManifest.xml. I gave it a quick try and it seems to have 7-zip installed in order to unpack it.

freddydk:
My bad on the $version - you are right.
Must be tired - will make another pass on this tomorrow.

freddydk:
I have checked in a fix, which uses the info in app.json when requesting symbols and uses the info from published apps to give the file name. It will also find the app version number in the database if an app isn't published.
And finally - new-navcontainer will unpublish symbols when running hybrid development.
All of this should fix this issue.

PR #356, opened 2019-02-20
improve bacpac export

ChrisBlankDe:
pass-though KeepUserData parameter.
fix error at bacpac export when multible versions of Data-tier Application Framework are installed. This problem can occure since commit/303395d0d0848f7b890f9620013f71849baffcb3 an can recur when changing the DACFx Version again

### DIFF
--- a/Bacpac/Export-NavContainerDatabasesAsBacpac.ps1
+++ b/Bacpac/Export-NavContainerDatabasesAsBacpac.ps1
@@ -20,7 +20,7 @@
  .Example
   Export-NavContainerDatabasesAsBacpac -containerName test -bacpacfolder "c:\demo" -sqlCredential <sqlCredential>
  .Example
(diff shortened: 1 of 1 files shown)


freddydk:
On the keepuserdata - I don't think that works. I tried to expose it, but it didn't really give me what I wanted (cannot remember the details). Have you tried it?

ChrisBlankDe:
sorry you haven't heard from me.
you were right: KeepUserData doesn't work as cleanly as I thought. I've now undone these changes.
All that remains is the version filter for the sqlpackage.exe which becomes necessary again with the update to 0.6.0.8 or newer if a container was created before the update and a bacpac with an older version was already created.

freddydk:
Chris, I have taken your fix and moved it to my dev branch and it will be part of the next version.
Thanks

Issue #374, opened 2019-03-07
New-NavContainer.ps1 fails with imported Database Backup


megel:
I use a database backup (including an App and C/Side changes) to create a new container. I use `SetupNavContainer.ps1` with a modification to Import the database backup:
```PowerShell
#>> Download the DB-Backup file ####
    # Use the DB-Backup file
    if ((Test-Path "c:\myfolder\database.bak" -PathType Leaf) -and ("$sqlServerType" -eq "SQLExpress")) {
        Log "Importing c:\myfolder\database.bak to container"
        $additionalParameters += @('--env bakfile="c:\run\my\database.bak"')
        $addUserToDatabase = $true
    }
#<< ####
```
![image](https://user-images.githubusercontent.com/15800069/53967720-06b26d80-40f6-11e9-80d6-6f0f8e23677e.png)

The scripts fails at `New-NavContainer.ps1`:
![image](https://user-images.githubusercontent.com/15800069/53967748-12059900-40f6-11e9-8ba4-5844332dd026.png)
`$enableSymbolLoading` is true (based on the ARM template `GetBCExt`)

The script works, when I change `New-NavContainer.ps1` to:
```PowerShell
    if ($enableSymbolLoading) {
        # Unpublish symbols when running hybrid development
        Invoke-ScriptInNavContainer -containerName $containerName -scriptblock {
            # Unpublish only, when App is present
            if ((Get-NavAppInfo -ServerInstance NAV -Name "Application" -Publisher "Microsoft") -ne $null) {
                Unpublish-NavApp -ServerInstance NAV -Name "Application" -Publisher "Microsoft"
            }
            # Unpublish only, when App is present
            if ((Get-NavAppInfo -ServerInstance NAV -Name "Test" -Publisher "Microsoft") -ne $null) {
                Unpublish-NavApp -ServerInstance NAV -Name "Test" -Publisher "Microsoft" 
            }
        } 
    }
```

freddydk:
Does that work???
Get-NavAppInfo will return $null unless you add -SymbolsOnly

The right way would probably be:

```
Get-NavAppInfo -ServerInstance NAV -Name "Application" -Publisher "Microsoft" -SymbolsOnly | Unpublish-NavApp
```

megel:
Hi hi sure 😂 in a database without these symbols ... Shame on me! I didn't check the opposite way 😳

megel:
I'll check tomorrow

megel:
[...]

PR #375, opened 2019-03-07
Check, the apps before unpublish

megel:
Fix for https://github.com/Microsoft/navcontainerhelper/issues/374


megel:
@freddydk I have included the parameter. Unfortunately I can not check, because I have no `Microsoft` Symbols in the container. But the New-NavContainer call does not fail including a database backup :)

PR #397, opened 2019-03-26
Invoke-ScriptInNAVContainer fix - use base64 for arguments

cmuris:
There was a problem when using Invoke-ScriptInNavContainer in a non-admin session. If one of the params contained apostrophes (i.e. $param = "UPDATE xxx SET `"Field A`" = '' " ), the temporary script file that was created contained a xml string with unescaped apostrophes. When running the temporary script file the xml string wasn't completly interpreted, resulting in an missing apostrophe in the param. This is now fixed by converting the whole XML to base64, which is converted back before the xml string is loaded.


freddydk:
Looks like a good fix, thanks

PR #425, opened 2019-04-17
Fix container variable name in Create-AlProjectFolderFromNavContainer

jose-lloret:
Container variable name in script was $imageName that did not exist. Fixed variable name to $containerName


freddydk:
Thanks for the catch.

jose-lloret:
You're welcome

PR #430, opened 2019-04-23
Ignore WARNING lines when getting companies

jimmymcp:
The output of Get-CompanyInNavContainer includes any warning lines that are returned by PowerShell e.g. licence about to expire, licence not compatible.

Added a Where to remove those lines. I guess this may be a problem for other commands as well but I'm not sure how you'd filter them out without changing each place Invoke-ScriptInNavContainer is called from.


freddydk:
A better fix is probably to add 

 | Where-Object { $_ -isnot [System.String] }

freddydk:
Thanks

PR #459, opened 2019-05-22
Add EventMessageFile for default NAV Server Instance

ChrisBlankDe:
Get rid of this message in the eventlog after running Get-NavContainerEventLog:

> The description for Event ID ### from source MicrosoftDynamicsNavServer$NAV cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.
> 
> If the event originated on another computer, the display information had to be saved with the event.
> 
> The following information was included with the event: 
> [...]

### DIFF
--- a/NavContainerHelper.psm1
+++ b/NavContainerHelper.psm1
@@ -32,6 +32,14 @@ if (!(Test-Path -Path $extensionsFolder -PathType Container)) {
     }
 }
 
+if ($isAdministrator) {
+    $RegPath = 'HKLM:\SYSTEM\CurrentControlSet\Services\EventLog\Application\MicrosoftDynamicsNavServer$NAV'
+    if (-not(Test-Path -Path $RegPath)){
+        New-Item -Path $RegPath -Force | Out-Null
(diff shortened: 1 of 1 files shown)


ChrisBlankDe:
Found some similar code here: [New-NavContainer.ps1 L827](https://github.com/microsoft/navcontainerhelper/blob/master/ContainerHandling/New-NavContainer.ps1#L827)
Why you running these on every container creation instead of module loading?

freddydk:
The things you found inside new-navcontainer is registering these things inside the container.
You are suggesting to add them on the host???

ChrisBlankDe:
Yes, i suggest them on the host.
Else, Every time you use Get-NavContainerEventLog you get the info that the source is unknown:
image

freddydk:
And how does this look with the change?

ChrisBlankDe:
After setting the registry key it looks like this (didn't find the server from first screenshot):
image

Its absolutly the same message but without the red marked header. Looks like on an regular onPrem installation.

freddydk:
This doesn't seem to be a problem with the latest BC versions

PR #470, opened 2019-05-28
Publish navcontainerapp

makuhn:
Publish-NavContainerApp:
- option ForceSync added to parameter SyncMode
- new parameter language added


freddydk:
The way to add a parameter is to add it as optional (as you have done) - but then only transfer the parameter to the underlying function if the parameter was set.
Pattern can be seen in Publish-NavContainerApp

            $syncArgs = @{}
            if ($syncMode) {
                $syncArgs += @{ "Mode" = $syncMode }
            }

makuhn:
I changed the script according to your example.
```
            $languageArgs = @{}
            if ($language) {
                $languageArgs += @{ "Language" = $language }
            }
    
            if ($install) {
                Write-Host "Installing $appName on tenant $tenant"
                Install-NavApp -ServerInstance $ServerInstance -Name $appName -Version $appVersion -Tenant $tenant $languageArgs
            }
```
But now I get following error:
_Install-NAVApp : A positional parameter cannot be found that accepts argument
'System.Collections.Hashtable'._

Do you have any idea why this fails? I cannot see any difference to the `$syncArgs` parameter.

freddydk:
use @languageArgs in the call

makuhn:
Thanks. I overlooked this. Now it works.

PR #485, opened 2019-06-06
switch landing page to /landing instead of :8180

tfenster:
Instead of using the unusual port of 8180 which lessens discoverability and increases complexity, use /landing on standard port 443. Works in connection with https://github.com/microsoft/nav-arm-templates/pull/109


freddydk:
Grabbed the changes to my dev branch manually - will close this.

PR #499, opened 2019-06-17
Update Get-NavContainerAppRuntimePackage.ps1

gunnargestsson:
App Name can contain unwanted file name characters


freddydk:
I think there are a lot of places where this is an issue. Why wouldn't you just add the appfile parameter yourself?

gunnargestsson:
I could, just wanted to contribute...  Your call

gunnargestsson:
This is the only place I have found that gives me an error with my AppSource App.  Would appreciate if this can be merged.  Can't see that this will cause any problems.

freddydk:
Ok

PR #529, opened 2019-07-26
error message fixed

### LINKED ISSUE #528: Generate-SymbolsInNavContainer raises unexpected
image

megel:
#528 ... fix the error message from call


freddydk:
Found the same problem - will use ([system.String]::Join("`n", $result)) instead (to get line breaks)

megel:
The change showed the real problem:

```code
[22922070] The following SQL Server error or errors occurred:  40607,"HY000",[Microsoft][SQL Server Native Client 11.0][SQL Server]Windows logins are not supported in this version of SQL Server.
at <ScriptBlock>, <No file>: line 24
```

#528 `Generate-SymbolsInNavContainer` does not support AzureSQL with configured container Authentication for DB (User/Password)

freddydk:
Thanks

PR #564, opened 2019-08-17
Enhancements for AL app handling

makuhn:
Create-AlProjectfolderFromNavContainer
- new parameter **useBaseAppProperties** added
- if set properties for app.json are retrieved from base app actually installed inside container

Publish-NewApplicationToNavContainer
- new parameter **saveData** added
- if set database is cleaned without deleting data of installed apps

New-NavContainer:
- Translations folder added when extracting files from base app in order to keep localization when building new base app


freddydk:
Thanks Martin, some of these things were on my list of things to look into.

Issue #617, opened 2019-09-13
"ClientSession is Uninitialized" when calling Run-TestsInNavContainer


jwikman:
This is similar to issue #527.

I'm calling Run-TestsInNavContainer from our pipelines, it has been working fine for some time now. But today I started to get "ClientSession is Uninitialized" when calling Run-TestsInNavContainer.

I'm using navcontainerhelper 0.6.4.1 (pipeline updates to latest version)

The target is a BC14 database.
Output when creating container:
```
Create CI-bld13 from mcr.microsoft.com/businesscentral/sandbox:w1
NavContainerHelper is version 0.6.4.1
Host is Microsoft Windows Server 2019 Standard - ltsc2019
Docker Client Version is 19.03.1
Docker Server Version is 19.03.1
Using image mcr.microsoft.com/businesscentral/sandbox:w1-ltsc2019
Disabling Health Check (always report healthy)
Creating Container CI-bld13
Version: 14.4.35602.0-W1
Style: sandbox
Platform: 14.0.35570.0
Generic Tag: 0.0.9.92
Container OS Version: 10.0.17763.678 (ltsc2019)
Host OS Version: 10.0.17763.720 (ltsc2019)
WARNING: The container operating system matches the host operating system, but the revision is different.
If you encounter issues, you might want to specify -isolation hyperv
Using locale en-US
Using process isolation
Disabling the standard eventlog dump to container log every 2 seconds (use -dumpEventLog to enable)
Using license file ***
Downloading C:\ProgramData\NavContainerHelper\Extensions\CI-bld13\my\license.flf
Files in C:\ProgramData\NavContainerHelper\Extensions\CI-bld13\my:
- AdditionalOutput.ps1
- AdditionalSetup.ps1
- license.flf
- MainLoop.ps1
- updatehosts.ps1
Creating container CI-bld13 from image mcr.microsoft.com/businesscentral/sandbox:w1-ltsc2019
92ecc421f71442c7a2186aefffd5c7663ab6a2c578c0d1c3c9a03da801e067bd
Waiting for container CI-bld13 to be ready
Initializing...
Starting Container
Hostname is CI-bld13
PublicDnsName is CI-bld13
Using NavUserPassword Authentication
Starting Local SQL Server
Starting Internet Information Server
Creating Self Signed Certificate
Self Signed Certificate Thumbprint D8F83C3DA6F06F736305706E9733546F89369884
Modifying Service Tier Config File with Instance Specific Settings
[...]

PR #620, opened 2019-09-13
Check that WebClientPort and FileSharePort are different

jwikman:
Prevent the messup in #617 to happen again :)

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -189,7 +189,7 @@ function New-NavContainer {
         $myScripts = @(),
         [string] $TimeZoneId = $null,
         [int] $WebClientPort,
-        [int] $FileSharePort,
+        [int] $FileSharePort = 8080,
         [int] $ManagementServicesPort,
         [int] $ClientServicesPort,
         [int] $SoapServicesPort,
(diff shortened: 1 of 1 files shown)


freddydk:
I guess run-tests and get-tests would also have to use the webclientport right?

freddydk:
This fix would cause FileSharePort to always be set (due to the defaulting) - not sure there would be any problem with that, but it is a difference.
Examle - if somebody is using additionalparameters to set the fileshareport - that script would now fail.

jwikman:
That's true, maybe that was my real problem yesterday? That the test used port 80, no matter what port I configured for the webclient? I couldn't find any relevant references to port 80, where is this set for the test functions?

Aah, now I understand why there were no default values for the parameters. I'll update the PR so the tests works without defaulting the parameter. 

But I guess that it'll still be an issue with additionalparameters. They'll be a bit of a challenge to test,  even if not impossible. Do you think we need to take that into consideration? I guess we have to...

freddydk:
A fix like this in SetupVariables:
```
if ("$fileSharePort" -eq "")          { 
    if ($webClientPort -eq "8080") {
        $httpSite = "N"
    }
    else {
        $fileSharePort = "8080"
    }
}
```
would mean that the container would switch off the fileshare if you use webclientport = 8080 without setting the fileshareport.
Beside that - run and gettests should use the webclient port.
will fix these things.

jwikman:
Yes, that is a better solution. 👍

freddydk:
But thanks for finding this

PR #648, opened 2019-10-05
Fix encoding for local special chars (mutated vowel, umlauts)

ChrisBlankDe:
This change allows to do request with special characters like ä, ö, ü in the body.

Without this change the following command will fail:
```
$packageId = "Änderungsprotokoll"
Invoke-BCContainerApi -containerName $containerName `
                      -CompanyId $CompanyId `
                      -credential $credential `
                      -APIPublisher "microsoft" `
                      -APIGroup "automation" `
                      -APIVersion "v1.0" `
                      -Method "POST" `
                      -Query "configurationPackages"  -tenant default `
                      -body @{ "code" = $packageId; "packageName" = $packageId; } | Out-Null
```

Error:
```
Unable to translate bytes[*] at index * from specified code page to unicode.
```

### DIFF
--- a/Api/Invoke-NavContainerApi.ps1
+++ b/Api/Invoke-NavContainerApi.ps1
@@ -145,7 +145,7 @@ function Invoke-NavContainerApi {
     }
     
     if ($body) {
-        $parameters += @{ "body" = $body | ConvertTo-Json }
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks will be included in the next version

freddydk:
Due to a regression - I will publish 0.6.4.7 asap, where this is included

freddydk:
Shipped

PR #685, opened 2019-10-23
exit function with 'return' instead of 'exit'

ChrisBlankDe:
The given 'exit' will couse the ISE to quit and the user will not be able to see the message that traefik is already initialized.
Maybe we should also change the Write-Host into a Write-Warning. What do you think?


freddydk:
In general - I want to switch to Write-Warning and Write-Information - but that change has a chance of being disruptive.
I am going to setup the containerhelper with pre-release versions which is supported by the gallery so that we can do these things and ship new versions less frequently.
Right now, the main focus has been to make CI/CD pipelines work - test running still has a few hickups - as soon as that is done, my focus will be to catch up on tests and create a pre-release version setup.
Make sense?

PR #686, opened 2019-10-23
Allow recreation of traefik container and configuration

ChrisBlankDe:


freddydk:
Thanks - was thinking about this exact scenario a while back when my traefik stopped working due to some misconfiguration.

PR #758, opened 2019-11-25
implement https support with the option to force http

tfenster:
In order to solve the problem that the mobile apps and modern Windows client can't connect to containers behind Traefik as they get mixed http / https traffic because the WebClient seems to serve some resources with hardcoded http, this change let's the containers run with SSL by default. It consists of three parts:

1. A new traefik template file which allows connections to endpoints with self-signed SSL certificates as the containers will use those: `insecureSkipVerify = true`
2. Changes to `New-NAVContainer` which enable SSL by default when running with `-useTraefik` and do the necessary setup for the Traefik rules: protocol is https and port is 443. The download site for the AL vsix file is always http, so the protocol for that is http as well. The parameter `-forceHttpWithTraefik` reverts those changes and the cmdlet will behave as previously
3. Changes to `Setup-TraefikContainerForNavContainers.ps1` which reference the new traefik template file. The parameter `-forceHttpWithTraefik` again reverts those changes and the cmdlet will behave as previously

I know it is documented somewhere, but I can't find it: How can I test the ARM templates with different navcontainerhelper sources?


freddydk:
Thanks, a few questions?

It looks like we are changing default behavior to use SSL between traefik and container - right? I think that is fine, but people with an existing Azure VM and updating containerhelper won't work anymore as I see it unless they specify -forceHttpWithTraefik in all their new-navcontainer calls.
(this will break all my CI/CD pipelines:-))

Maybe we could remove the forceHttpWithTraefik parameter in new-navcontainer and set it automatically (where we check that the traefik container is setup with usetraefik) based on how the traefik container is setup?

Also it seams like the traefik.toml config file will always be names _https unless you specify the file yourself. Shouldn't this be defaulted based on -forceHttpWithTraefik? (then it would be easier to default the forceHttpWithTraefik in New-NavContainer).

tfenster:
> It looks like we are changing default behavior to use SSL between traefik and container - right? I think that is fine, but people with an existing Azure VM and updating containerhelper won't work anymore as I see it unless they specify -forceHttpWithTraefik in all their new-navcontainer calls.
> (this will break all my CI/CD pipelines:-))
I have to admit, I didn't think about the existing VMs... :(

> Maybe we could remove the forceHttpWithTraefik parameter in new-navcontainer and set it automatically (where we check that the traefik container is setup with usetraefik) based on how the traefik container is setup?
Very good idea, thanks! I removed the param and tried to check the config file [here](https://github.com/microsoft/navcontainerhelper/pull/758/files#diff-98e8e0a0fad1ad04d5c8df136c4f7f47R360-R364). Really not very elegant, but I couldn't come up with anything better
 
> Also it seams like the traefik.toml config file will always be names _https unless you specify the file yourself. Shouldn't this be defaulted based on -forceHttpWithTraefik? (then it would be easier to default the forceHttpWithTraefik in New-NavContainer).
Absolutely true. I misunderstood the parameter $traefikToml, but reverted that now

freddydk:
Agree, I don't like the parameter on New-NavContainer either.
Since we already have hardcoded the path for traefikforbc - couldn't we just check and see whether this file exists:
(Join-Path $traefikForBcBasePath "config\template_traefik_https.toml")
then it is https - else it is http.
Wouldn't that work?

tfenster:
Once more, I handled the traefiktoml param wrong. The file below (Join-Path $traefikForBcBasePath "config") will always be called traefik.toml as this is the config file name traefik picks up automatically. We also need to consider the case where someone brings in their one template file by setting e.g. $traefikToml to https://mydomain.com/traefik_template.toml, so I think the only reliable way is to check the content of the config file

freddydk:
http://aka.ms/getbc2 will use the dev branch in the navcontainerhelper repo.
http://aka.ms/getbc?nchbranch=https%3A%2F%2Ftheurlofyourrepo - should read containerhelper from that repo (haven't tried recently)

if the config file always is the same - you could move the code a little further down:

```
        $forceHttpWithTraefik = $false
        if ((Get-Content "$traefikForBcBasePaht\config\traefik.toml" | Foreach-Object { $_ -match "^insecureSkipVerify = true$" } ) -notcontains $true) {
            $forceHttpWithTraefik = $true
        }
```

and then remove the parameter:-)

tfenster:
Once more, you are very right... Not my best PR, to put it politely :)

freddydk:
you still need to replace the reference to the non-existing parameter with "$traefikForBcBasePath\config\traefik.toml"
right?

tfenster:
absolutely... 

A first quick test using the nchbranch param didn't work (seems to not have run any nch commands), but before I do more stupid things, I'll call it a day and try again tomorrow or on Monday ;)

freddydk:
Goodnight:-)

You are missing config\ in front of traefik.toml

and I checked - the nchbranch needs to be a direct download to a .zip file with the containerhelper - so:

http://aka.ms/getbc?nchbranch=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fnavcontainerhelper%2Farchive%2Fdev.zip 

should work.

tfenster:
Amazing how many simple things I am able to get wrong :)

Ok, tried again and tested the following:
- Pre-existing VM created from ARM template:
1. Downloaded my repo as .zip
2. Extracted it and ran `Import-Module .\NavContainerHelper.psm1`
3. Ran `New-NavContainer -accept_eula -containerName testnch -imageName mcr.microsoft.com/businesscentral/onprem:ltsc2019 -useTraefik -auth NavUserPassword`
4. As expected this creates a container with `useSSL=N` and traefik set up to use http as protocol and port 80:
``` 
            "Env": [
                ...
                "useSSL=N",
                ...
            ],
            ...
            "Labels": {
                ...
                "traefik.protocol": "http",
                ...
                "traefik.web.port": "80",
                ...
            }
```
5. Ran `Setup-TraefikContainerForBCContainers -PublicDnsName td19-traef.westeurope.cloudapp.azure.com -ContactEMailForLetsEncrypt tobias.fenster@cosmoconsult.com -Recreate`
6. Now the traefik.toml file has `insecureSkipVerify = true`
7. Ran `New-NavContainer -accept_eula -containerName testnch -imageName mcr.microsoft.com/businesscentral/onprem:ltsc2019 -useTraefik -auth NavUserPassword` again
8. This creates a container with `useSSL=Y` and traefik set up to use http and port 443:
```
            "Env": [
                ...
                "useSSL=Y",
                ...
            ],
            ...
            "Labels": {
                ...
                "traefik.protocol": "https",
                ...
                "traefik.web.port": "443",
                ...
            }
```
9. Connecting through the mobile app works and the browser debug console shows no "Mixed mode" warnings

- New VM created with http://aka.ms/getbc?nchbranch=https%3A%2F%2Fgithub.com%2Ftfenster%2Fnavcontainerhelper%2Farchive%2Fmaster.zip
1. Configuration shows the same as above under 8. and traefik.toml has `insecureSkipVerify = true`
2. The initialization fails with 
```
9:26:36 AM at , C:\demo\setupVm.ps1: line 112
9:26:36 AM The specified module 'navcontainerhelper' was not loaded because no valid module file was found in any module directory
```
3. I then changed to `C:\DEMO\navcontainerhelper-master` and ran `Import-Module .\NavContainerHelper.psm1` 
4. Creating a new container using `New-NavContainer -accept_eula -containerName newnch -imageName mcr.microsoft.com/businesscentral/onprem:ltsc2019 -useTraefik -auth NavUserPassword` works as expected with the same config as above under 8.
4. Connecting also works

What did I miss this time? :D

freddydk:
Nothing (I think) - will merge and run some test.

Thanks a million for your effort:-)

/Freddy

Issue #760, opened 2019-11-26
Import-TestToolkitToNavContainer fails on v15


epernst:
Hi Freddy,

I am getting an error when using Import-TestTookitToNavContainer (NCH 0.6.4.19 + Onprem:1910-CU1):

Failing error message: The tenant 'default' is not accessible, because it is in state Mounting. 
            at <ScriptBlock>, <No file>: line 29

Whenever I debug the code, then I have no problem, but when running our automated build it fails.

The reason is, as far as I can see, line 100 (Set-NavServerInstance $serverInstance -restart).
For a short moment the state is here "Mounting" (we are using a multi-tenant server), which causes the attempt to publish the test extensions to fail.

I have experienced the same issue (state Mounting) in other situations (mostly when mounting), but there I'm able to insert a "Wait Until" command, but I cannot see how I can do this with your function.

freddydk:
QQ - are you using process isolation?
I will insert a wait for tenant in the import function.

In the meantime, you could do this before calling the import function.
```
Invoke-scriptinNavContainer {
    $mockAssembliesPath = "C:\Test Assemblies\Mock Assemblies"
    if (Test-Path $mockAssembliesPath) {
        $serviceTierAddInsFolder = (Get-Item "C:\Program Files\Microsoft Dynamics NAV\*\Service\Add-ins").FullName
         if (!(Test-Path (Join-Path $serviceTierAddInsFolder "Mock Assemblies"))) {
             new-item -itemtype symboliclink -path $serviceTierAddInsFolder -name "Mock Assemblies" -value $mockAssembliesPath | Out-Null
            Set-NavServerInstance $serverInstance -restart
            Start-sleep -seconds 5
        }
    }
}
```

freddydk:
Updated the sample with a wait loop, waiting for not mounting

epernst:
Thanks, I actually did the same in a temporary version of it, except I also used -ForceRefresh, where the results are always correct. 
And while we are at the -ForceRefresh param, then it's a shame you don't include that one in Get-NavContainerTenant, I use it all the time.

freddydk:
I guess I can add that.

freddydk:
like 
                    while (Get-NavTenant $serverInstance -forcerefresh | Where-Object { $_.State -eq "Mounting" }) {
                        Start-Sleep -Seconds 1
                    }

freddydk:
Removed forcerefresh again as some of my tests on the containerhelper failed due to this.
[...]

PR #761, opened 2019-11-26
AppFile variable does not exists in context

epernst:
Just a small change.

When using the Sync-NavContainerApp function, then it currently writes "Synchronizing $appFile on $tenant", which results in either blank or potential wrong information.


epernst:
Didn't know additional pushes to my fork resulted in my pull request being updated. But then again why would it be different than with Azure DevOps.
Second push is merely what I referred to in https://github.com/microsoft/navcontainerhelper/issues/760#issue-528704851 issue. I have missed ForceRefresh in Get-NavContainerTenants.
And why limit the function to MultiTenant implementations, when its just as useful whenbit's not?
So I removed the error it not multi-tenant.

freddydk:
Are you sure that this works?
You do not have a Param($Params) in the invoke-script scriptblock. I would assume that the parameter won't get transferred.
Also no parameter description is included in the file.

epernst:
Oh yes, (param) was missing. Which is why you should not do refactoring right before submitting - the params was inside the invoke - but moving it outside it becomes easier to debug.
Also updated parameter docs.

freddydk:
Thanks Erik

Issue #765, opened 2019-11-27
Convert-AlcOutputToAzureDevOps fails on new AppSourceCop rules


jwikman:
Convert-AlcOutputToAzureDevOps fails for the new AppSourceCop warning 
`warning AS0055: The AppSourceCop configuration must specify the list of countries targeted by the application.`

This error is thrown:
`##[error]Convert-ALCOutputToAzureDevOps : Cannot bind argument to parameter 'Path' because it is null.
`

A PR is coming your way soon... ;-)

freddydk:
Fixed in 0.6.4.21

PR #766, opened 2019-11-27
Fix issue https://github.com/microsoft/navcontainerhelper/issues/765

jwikman:
`Test-Path` with a null Path parameter fails with error `Cannot bind argument to parameter 'Path' because it is null.`

I don't really understand why `Test-Path $Matches[3]` is used. 
But if you wan't to check if there where 2 or 3 matches, checking if `$Matches[3]` is $null should be ok.

This is tested in my pipeline for `warning AS0055: The AppSourceCop configuration must specify the list of countries targeted by the application.` and it seems to do it's thing.


freddydk:
Thanks

PR #775, opened 2019-12-05
get tenant database from tenant property if possible 

svengrav:
If the database name differs from the tenant name, then the script currently does not work. The change causes the database name to be taken from the tenant configuration.


freddydk:
Looks good, accepted.

freddydk:
Thanks

Issue #791, opened 2019-12-19
Network-related or instance-specific error on creating new Container with options -useNewDatabase or -useCleanDatabase


aho81:
When creating a new Container wie New-BCContainer with the parameters useNewDatabase or useCleanDatabase I get the following error:

A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
at <ScriptBlock>, <No file>: line 7
A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
At Z:\Resources\Repository\navcontainerhelper\ContainerHandling\Invoke-ScriptInNavContainer.ps1:37 char:13
+             Invoke-Command -Session $session -ScriptBlock $scriptbloc ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Invoke-Sqlcmd], SqlException
    + FullyQualifiedErrorId : SqlExectionError,Microsoft.SqlServer.Management.PowerShell.GetScriptCommand
    + PSComputerName        : 10e5fc537d175eb87cde2b83659aa5dcceb93468e16898aedad811b89ebcf0ee

freddydk:
shipped in 0.6.4.24

PR #792, opened 2019-12-19
Added DatabaseServerInstance information to Sql-Queries

aho81:


aho81:
PR should solve #791

freddydk:
Copy-NavDatabase also has two parameters - databaseserver and databaseinstance and should probably also be set if this is to work, right?

freddydk:
And Remove-NavDatabase (for removing the copied database:-))

aho81:
I'll have a look? Shall I provide those as different pull requests?

freddydk:
If you push more changes they will update the PR

Hent Outlook til iOS<https://aka.ms/o0ukef>
Fra: aho81 <notifications@github.com>
Sendt: Friday, December 20, 2019 1:32:59 PM
Til: microsoft/navcontainerhelper <navcontainerhelper@noreply.github.com>
Cc: Freddy Kristiansen <Freddy.Kristiansen@microsoft.com>; Comment <comment@noreply.github.com>
Emne: Re: [microsoft/navcontainerhelper] Added DatabaseServerInstance information to Sql-Queries (#792)


I'll have a look? Shall I provide those as different pull requests?

aho81:
I added the databaseserverinstance parameter also to sqlcmd calls in the Export-NavContainerDatabaseAsBacpac as I got the same error as in clean database. Also, I looked for all other calls to sqlcmd an they seem fine :-). 

Additionally, I added databaseserver / -instance properties to Copy-/Remove-NavDatabase calls in clean db but I'm not sure that those where necessary because I got no errors about these. 

I explicitly did not change the Copy-/Remove-NavDatabase calls in New-NavContainer as I'm not sure why this works without ;-). All other calls do have those parameters. 

Shall the calls in New-NavContainer have to be changed, too? I get no error when using this function, but in multitenant-scenarios there are 3 dbs; financials (appdb), and "tenant" and "default" (tenant-dbs).

freddydk:
Renaming parameters is not anything I like to do.
It means that any partner using this export-tobacpac function need to change their script based on the version of containerhelper.
I realize that the right name is serverinstance - but I would rather keep the name.

freddydk:
As I had done some changes to Clean-database in my dev branch (same areas) I have merged your changes into my dev branch (except for the parameter name changes - I like to only change the actual issues)
Thanks for the contribution.

freddydk:
If you discard your changes and synchronize then your changes should now be in master.
Building 0.6.4.24 now - publishing if all tests passes.

Sorry for doing the PR this way, that is normally not the right way, but I had to get the other change in - and I wanted to include your chnage,

aho81:
That’s ok. I’ll check out the new master when published.

I renamed the internal parameters to make their purpose more clear as you have different vars for databaseserver and databaseserverinstance, but this is not neccessary in a functional way.

What about the added parameters to Copy-/Remove-NAVDatabase? Did you take these?

freddydk:
Yes

freddydk:
shipped in 0.6.4.24

PR #805, opened 2020-01-04
Force in Install-NavContainerApp - Tenant in Get-NavContainerTenants

epernst:
We have had a number of situations where using Install-NavContainerApp in our pipelines have given "WARNING: UnhandledErrorMessage" failures followed by "Windows PowerShell is in NonInteractive mode. Read and Prompt functionality is not available.".
I found that Install-NavContainerApp did not already have a Force parameter (unlike Uninstall-NavContainerApp and most other functions), but that adding it solved the problem. 
Now we have a "local" function called Install-NavContainerAppWithForce, but would love to get rid of it again.

Also I must still use "| Where-Object -Property Id -eq $TenantId" when ever I use Get-NavContainerTenants and do not use ForceRefresh. While there is no reason to limit it. Get-NavTenant supports it with and without.


freddydk:
Not sure I understand the changes in Get-NavContainerTenants.
You set the tenant parameter to mandatory and you have two parameter names on that.
Then you check whether it is set - how does this work?

epernst:
Yes it works. :) 
I means that Tenant is mandatory when using the ForceRefreshTenantState parameterset, but otherwise optional.

freddydk:
Ahh, so this line:

`[Parameter(ParameterSetName = 'UseContainerName')]`

Should probably be

`[Parameter(Mandatory=$false, ParameterSetName = 'NoForceRefresh')]`

But then - if forcerefresh, the defaulting of tenant has gone, right? Isn't that a breaking change?

epernst:
No, there is no NoForceRefresh parameterset. 
And Mandatory=$false should never be specified, as that is the default vaule, so [Parameter(ParameterSetName = 'UseContainerName')] = [Parameter(Mandatory=$false, ParameterSetName = 'UseContainerName')].
Not sure it's possible to use a default value for one parameterset and a blank value for a different.

freddydk:
Got it - of course the default parameter set name should change as well - usecontainername just doesn't state anything and I know that mandatory false is default - I was just looking for ways that this wouldn't look like an error/mistake:-)

re your second comment - you could default the tenant in code if forcerefresh - and always have tenant as non-mandatory (in order to not break anything)

freddydk:
Wouldn't this work:

```
function Get-NavContainerTenants {
    [CmdletBinding(DefaultParameterSetName = 'UseContainerName')]
    Param (
        [Parameter(Mandatory = $false)]
        [string] $containerName = "navserver",

        [Parameter(Mandatory = $true, ParameterSetName = 'ForceRefreshTenantState')]
        [switch] $ForceRefresh,

        [string] $Tenant = "",

        [switch] $Force
    )
    $Params = @{ "Force" = $Force }
    If ($ForceRefresh) {
        $Params += @{
            "ForceRefresh" = $ForceRefresh
        }
        if (!($tenant)) {
            $tenant = "Default"
        }
    }
    if ($Tenant) {
        $Params += @{
            "Tenant"       = $Tenant
        }
    }
    Invoke-ScriptInNavContainer -containerName $containerName -ScriptBlock {
        Param( [PsCustomObject] $Params)
        Get-NavTenant -ServerInstance $ServerInstance @Params
    } -argumentList $Params
}
```

epernst:
It would not work, as we need Tenant to be mandatory when using ForceRefresh, but not when its not. 
Also you should be using [ValidateNotNullorEmpty()] for the Containername, this way you can use Mandatory=False and specify a value, but still make sure it is not a blank ("") value.

epernst:
I tested it, and it seems that it does work, with less risk of having a breaking change. So pushed a new update.

freddydk:
I agree I should use [ValidateNotNullorEmpty()] for the Containername in all functions.

one question:

Why is this:
`if (-not $tenant -or '' -eq $tenant) {`
better than
`if (!($tenant)) {`

?

epernst:
Because when you specify "" as the default, it is always set (which means that its true or not null). 
So we only want to override this if the value is "".
As to (!($tenant))  vs (-not $tenant) then it's just because the "-not" more readable than "!".

freddydk:
My point is that my PowerShell knowledge - this is how ! or -not works:

```
$test = ""
if (!($test)) {
    Write-Host "empty or null"
}
```

testing for empty afterwards doesn't add anything - or?

epernst:
Looks like you're right. I've always done both, just to be sure...

freddydk:
Thanks for your contribution

PR #822, opened 2020-01-21
Fix Invoke-ScriptInNavContainer when running unelevated

marknitek:
When running Invoke-ScriptInNavContainer unelevated and the scriptblock fails, the Invoke- Script does not throw any error. This causes for example the Publish-NavContainerApp to report success when in fact it failed. 

Simply checking the `$LASTEXITCODE` should solve the issue but maybe theres a bether solution? 
It seems like docker exec is writing errors to stout instead of sterr?


freddydk:
Thanks

PR #833, opened 2020-02-05
escape $ with `

Tino-T:
Pull Request inserts a missing escape character.


freddydk:
Right you are, thanks

PR #855, opened 2020-02-17
Fix illegal chars in app file name

haweg:
This modification prevents errors due to illegal characters in the app file name that occur when such characters are used in the app name.
Illegal characters are now simply removed from the app file name (which is in line with the behavior of the VSCode AL extension).


haweg:
Hello @freddydk , I just saw a little mistake I made in the Sort-AppFoldersByDependencies.ps1 (line 96 & 97). Is it best when I open a new pull request for this?

freddydk:
Yes please, this one is closed, thanks for the help btw,

PR #856, opened 2020-02-18
fixed array of unknown app dependencies

haweg:
In addition to PR #855 - fixed array of names of the unkown app dependencies in Sort-AppFoldersByDependencies.ps1


freddydk:
Ahh - you moved the appfilename outside of the loop, thanks for catching.

Issue #897, opened 2020-03-18
New-NavContainer run into a with option IncludeCSide and added AdditionalSetup.ps1 by folders


megel:
**NavContainerHelper v0.6.5**

```powershell
$imageName   = "microsoft/dynamics-nav:2016-cu13-de"
$licenseFile = "..."
New-NavContainer nav2016 -isolation hyperv -imageName $imageName -accept_eula -alwaysPull -useBestContainerOS -accept_outdated -updateHosts -licenseFile "$licenseFile" -auth UserPassword -dns "8.8.8.8" -Credential ([PSCredential]::new("username", (ConvertTo-SecureString -String "password" -AsPlainText -Force))) -additionalParameters $additionalParameters -includeCSide -doNotExportObjectsToText
```

An *very* unexpected error occured 😥:

![image](https://user-images.githubusercontent.com/15800069/77002808-6a53b580-695c-11ea-8fd5-4315f6a3f2f2.png)

I found out, that the call `(Get-Item "$programFilesFolder\*\RoleTailored Client")` returned $null:

![image](https://user-images.githubusercontent.com/15800069/77002153-53f92a00-695b-11ea-928f-1f2c80db412f.png)

![image](https://user-images.githubusercontent.com/15800069/77002434-c9fd9100-695b-11ea-89f7-c817a9703b9c.png)

I have inserted this code:
![image](https://user-images.githubusercontent.com/15800069/77002556-00d3a700-695c-11ea-9819-68408f664d24.png)

Result:
![image](https://user-images.githubusercontent.com/15800069/77002591-0b8e3c00-695c-11ea-8790-7ce1a943a63f.png)

@freddydk, pull request will follow tomorrow.

megel:
The problem might be the added environment parameter `folders` for the container start.

The container start contains a parameter:
`-e folders=c:\run\my=https://...blob.../setup.zip`

My Zip folder is extracted into `c:\run\my` and contains:

```cmd
- AdditionalSetup.ps1
- ... some other stuff
```
First of all, `New-NAVContainer -includeCSide ...` which add some code to the AdditionalSetup.ps1. The added code causes to copy the RTC folder to the container-folder under `c:\ProgramData\...\my`.

Unfortunately the added `-e folder=...` seems to overwrite the created `AdditionalSetup.ps1` in the `c:\run\my` folder of the container 😣

megel:
The problem might be the following code in `New-NAVContainer.ps1`:
[...]

PR #898, opened 2020-03-18
#897 Ensure, that the RTC folder exists

megel:
This fixed the problem with NAV2016 on my test environment. I have unfortunately no idea, why the folder did not exist.


megel:
@freddydk

PR #904, opened 2020-03-20
First check "Docker is running" before checking "Windows Container"

Tino-T:
If Docker is not running, you will get the error message that you should switch to Windows containers:

image

After this pull request, the error message that Docker is not running is displayed:

image


freddydk:
Thanks

Issue #933, opened 2020-04-04
"The variable '$appsFolder'cannot be retrieved because it has not been set." error in Publish-NewApplicationToNavContainer


eclipses:
Hi Freddy,

Yesterday I noted that you created the function **Publish-NewApplicationToNavContainer** so I was trying to use it to install an app as a prerequisite after the creation of a BC container in our CI/CD pipelines.
I'm receiving multiple errors because of a variable _$appsFolder_ not set, I've noted that it is assigned only inside a condition
`if ($restoreApps -ne "No") {`
but it used out of the condition.
I've tried also forcing the code to go inside the condition, passing the parameter _-restoreApps Yes_ , but in the end, the publication of my app still fails with 
`Status Code 422 : Unprocessable Entity`
I found in old issues that you suggested to increase the MemoryLimit but I'm NOT using hyperv.
Please when you have time have a look of the code and test this function.

Thanks
Sergio Castelluccio

eclipses:
I have an important update, I opened the Event Viewer inside the container after received the error
`Status Code 422 : Unprocessable Entity`
and I found that the issue is that during the publication of the app, the compilation fails because it cannot find the DLLs passed with the parameter _-appDotNetPackagesFolder_. 
I tried to restart the container and then run the same command than it worked.
I have a feeling that after importation of the DLLs the function should restart the BC Instance in order to enable access to them. 
This is the result of my test, please let me know. I'm importing an App with dependency in two DLLs.

eclipses:
Hi Freddy,

I'm trying to figure out if there is a way to include a copy of two dlls and/or installation of an ISV app during the creation of a BC Container. 
Maybe I could use AdditionalSetup.ps1 but I did not find yet at which stage copy the files inside the container. Any suggestion if this can be achieved with the current version of navcontainerhelper or maybe could be integrated in the future?

freddydk:
If you have a folder with add-ins, you can share that to the add-ins folder under c:\run container using:
-additionalparameters @('--volume c:\myaddinsfolder:c:\run\add-ins')
The content of that folder will be copied to the corresponding add-ins folders during startup of the container.
see more here:
https://github.com/microsoft/nav-docker/blob/master/generic/Run/SetupAddIns.ps1

freddydk:
Shipped in 0.6.5.6

PR #934, opened 2020-04-05
Restart instance after creation of a new symbolic link and $appsFolde…

eclipses:
Restart instance after the creation of a new symbolic link and $appsFolder out of condition. This should fix the issue #933


freddydk:
Thanks.
Did you test this with both a multitenant container (with multiple tenants - at least 2) and a single tenant container?

eclipses:
I tested it in a single tenant, give me 10 minutes so that I will test in a mutitenant

eclipses:
Now, tested on a multitenant (2 tenants) passed as well

PR #935, opened 2020-04-06
Use win32 version of the alc.exe when compiling apps 

RonKoppelaar:
The win32 version of the alc.exe is based on .Net core and improves performance significantly.


freddydk:
Already had the same in my dev branch - but only one of the changes, as the other change probably won't work with PowerShell.
Thanks for the PR though - the other one is in master now and will likely publish tomorrow.

RonKoppelaar:
Thx!

Met vriendelijke groet,
Ron Koppelaar
Architect

Issue #944, opened 2020-04-15
Get-TestsFromBCContainer - The property 'Tests' cannot be found on this object.


sirhc101:
Hi freddy,

hopefully you can help me with a problem. I've up and running 10+ build pipelines, based on your great blog posts regarding CI/CD. The pipelines all are working very well and running on a self hosted DevOps agent.
Now we started with a new product and cloned the scripts, created the build pipelines and got in trouble. Everything works fine, until it comes to step "Run Tests". There I receive:
```cmd
09:01:49.2346238Z ##[debug] Arguments: '-NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'C:\agent\_work\_temp\40182dc3-14b0-4eea-9845-12df4ce3a02c.ps1'"'
09:01:49.2359248Z ##[debug] FileName: 'C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe'
09:01:49.2372068Z ##[debug] WorkingDirectory: 'C:\agent\_work\24\s'
09:01:49.2408445Z ##[command]"C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'C:\agent\_work\_temp\40182dc3-14b0-4eea-9845-12df4ce3a02c.ps1'"
09:02:00.6635002Z Open form
09:02:19.9560227Z Setting Extension Id 79081a2a-d3db-4c71-b144-af22b403e2e4
09:02:37.8715012Z Title: Command Line Test Tool Edit - Command Line Test Tool - DEFAULT
09:02:37.8735991Z Title: {0001fd97-0000-0001-0c00-0000836bd2d2}
09:02:37.8882183Z [
09:02:37.8886339Z     {
09:02:37.8890326Z         "type":  "ClientGroupControl",
09:02:37.8894295Z         "mappingHint":  "TOOLBAR",
09:02:37.8898320Z         "name":  "CommandBar",
09:02:37.8902239Z         "children":  [
09:02:37.8906298Z                          {
09:02:37.8910396Z                              "type":  "ClientGroupControl",
09:02:37.8914408Z                              "mappingHint":  "",
09:02:37.8918417Z                              "name":  "Control_Home",
09:02:37.8922817Z                              "children":  [
09:02:37.8926928Z                                               {
09:02:37.8931027Z                                                   "type":  "ClientGroupControl",
[...]

PR #947, opened 2020-04-16
Bugfix PsTestFunctions.ps1, Function Get-Test

sirhc101:
As described in Issue #944 there is a problem in PsTestFunctions.ps1 in handling the repeater offset correctly. The problem seems only to occur if more than 15 tests are existing.

With the code modifications I was able to run pipelines with more and less than 15 tests included. Tested with 12 build pipelines and BC16 (sandbox).

The console output has been changed to this:
image
image

### DIFF
--- a/AppHandling/PsTestFunctions.ps1
+++ b/AppHandling/PsTestFunctions.ps1
@@ -204,7 +204,7 @@ function Get-Tests {
 				Write-Debug $("[Row Index: " + $rowIndex + " (Index: " + $index + ", Offset: " + $repeaterOffset + ") ] Nothing found.")
 			}
 			
-			if ($rowIndex -ge ($originalRepeaterOffset - 1))
+			if (($rowIndex -ge ($originalRepeaterOffset - 1)) -and ($originalRepeaterOffset -ne 0))
 			{
 				Write-Debug "+1 Scroll Repeater ."
 				$clientContext.ScrollRepeater($repeater, 1)
@@ -214,6 +214,7 @@ function Get-Tests {
 	}
 
     $index = 0
+	$repeaterCount = [int]$repeater.DefaultViewport.Count
 	if ($repeaterOffset -gt 0) {
 		$repeaterCount = $repeaterCount + $repeaterOffset
 		$repeaterOffset = 0
@@ -274,7 +275,7 @@ function Get-Tests {
             }
 		}
 		
-		if ($rowIndex -ge ($originalRepeaterOffset - 1))
+		if (($rowIndex -ge ($originalRepeaterOffset - 1)) -and ($originalRepeaterOffset -ne 0))
 		{
 			$clientContext.ScrollRepeater($repeater, 1)
 			$repeaterOffset = $repeaterOffset + $originalRepeaterOffset
--- a/AppHandling/PsTestFunctions.ps1
+++ b/AppHandling/PsTestFunctions.ps1
@@ -167,25 +167,68 @@ function Get-Tests {
 
     $repeater = $clientContext.GetControlByType($form, [ClientRepeaterControl])
     $index = 0
-    $repeaterCount = [int]$repeater.DefaultViewport.Count
+	$repeaterCount = [int]$repeater.DefaultViewport.Count
 	$repeaterOffset = [int]$repeater.Offset
 	$originalRepeaterOffset = [int]$repeater.Offset
+	if ($debugMode) {	
+		Write-Debug "No. of Records: $repeaterCount"
+		Write-Debug "Repeater Offset: $repeaterOffset"
+	}
+	if ($repeaterOffset -gt 0) {
+		$repeaterCount = $repeaterCount + $repeaterOffset
+		$repeaterOffset = 0
+		Write-Debug "-1 Scroll Repeater."
+		$clientContext.ScrollRepeater($repeater, -1)		
+	}
+	
+	if ($debugMode) {	
+		Write-Debug "Iterate through Repeater control..."	
+		while ($true)
+		{
+			$rowIndex = $index - $repeaterOffset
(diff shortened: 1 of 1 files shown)


freddydk:
I am trying to figure out what exactly your fix is doing.
It looks like you are looping through the repeater first to "fill it up" and then continue into more or less the same code as before.
Is this the strategy?

sirhc101:
More or less, yes! The first loop is basically for debugging purpose.
The real "fix" is regarding the repeater offset. Sometimes (I cannot look into the DLL so I don't know the reason) the repeater is starting with an offset and you need to scroll back (-1) to make sure, that you find the codeunit object and the corresponding tests.

The big problem is, that not everytime something goes wrong an error occur, because if the scripts has found the codeunit object everything is fine. But in real this can cause to not executed tests.

It's really a though problem and really hard to reproduce. Would definitly be helpful to look into the DLL file providing the repeater control.

freddydk:
Could you either here or on the issue paste the script used (I had multiple versions of the run-tests)

Thanks

freddydk:
Looking at this, I am afraid of side effects of this change.
Earlier changes have taught me, that I should only fix exactly what's wrong and not reintroduce variables, move code and other things.
I do however realize that the issue you have is real and need a fix.
Your description that the repeater offset is greater than 0 coming into the get-tests is definitely a problem and I am investigating why that can happen and I have sent an email to the client team (responsible for the DLLs) what the best way of ensuring that the offset is 0.

freddydk:
Would it be possible for you to try out a new PSTestFunctions and ClientContext from here:
https://raw.githubusercontent.com/microsoft/navcontainerhelper/dev/AppHandling/ClientContext.ps1
https://raw.githubusercontent.com/microsoft/navcontainerhelper/dev/AppHandling/PsTestFunctions.ps1
It does an implicit select first row every time it starts.

sirhc101:
Yes, I will test it within my pipelines.

I will come back to you with the results.

sirhc101:
Unfortunately this does not work
image

freddydk:
I do realize that you probably went through a lot of this testing already, but I need to see what actually happens on your box, so I have added a xx-load of debug info in get-tests - enough to ensure that I should be able to deduct what happens.
By running this myself, I have an idea of what might happen - if you can re-run and send me the complete output - then I will know more.
Thanks for your help.

sirhc101:
To be honest, my pipelines are doing the most of the work ;-)

I updated Get-TestsFromNavContainer.ps1 and PsTestFunctions.ps1 from your last commit (c36c1ca0826b5eb9b17f56b730a4aff7c52583d8) and my pipelines are running now... this will take some time, but I will paste the results

sirhc101:
Here are the results from the first failing pipeline:
```
Formatted command: . 'C:\agent\_work\25\s\scripts\Run-Tests.ps1' -appFolders "test,app"
========================== Starting Command Output ===========================
"C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'C:\agent\_work\_temp\0d04d5c2-e42f-4d89-9540-4b8e9f293712.ps1'"
Open page 130455
Setting Extension Id e4664942-0212-41a2-aa2d-cf376adb84a6
Offset: 0
Offset: 0
Index:  0, Offset: 0, Count:  2
Row - lineType = 1, run = Yes, CodeunitId = 0, codeunitName = '', name = ''
Index:  1, Offset: 0, Count:  2
Row - lineType = 1, run = , CodeunitId = , codeunitName = '', name = ''
Index:  2, Offset: 0, Count:  2
Scroll
Index:  2, Offset: 0, Count:  2
Breaking - rowIndex: 2
Open page 130455
Setting Extension Id 79081a2a-d3db-4c71-b144-af22b403e2e4
Offset: 15
Offset: 0
Index:  0, Offset: 0, Count:  15
Row - lineType = 2, run = Yes, CodeunitId = 62030, codeunitName = '', name = 'bdev.Address Validation Test'
Add test CODEUNIT NAME
Title: Command Line Test Tool Edit - Command Line Test Tool - DEFAULT
[...]
```
Some context information. The Test Codeunit which was found contains 17 tests. The offset (15) is resulting in finding only two or three of them. Last time I debugged the codeunit wasn't found as first element, but this is expected by your script to append the tests to the codeunit.

Hope that helps!

If it might help, we can look at our server together and you can play around!?

freddydk:
One thing I see now is, that you have appfolders = test,app - and it tries to add tests for app first and then test.
Could you modify the pipeline to only run tests in the testfolders - this will tell me if this is caused by adding an empty test app first.

freddydk:
I created a small project with a pipeline and all did the same as you (even have 17 tests in my test project - still no repro).
What docker image are you using? (which BC version)

freddydk:
For the first time, I now have a repro of the offset being 15...
image
My pipeline isn't failing because my SelectFirstRow fix fixes the issue in my pipeline.
Will try to investigate why this happens.

freddydk:
When I removed selectfirstrow, I got the same issue as you.
I have made another fix - thinking that the selectfirstrow might not work correctly in all versions.
Could you download a new pstestfunctions from here:
https://raw.githubusercontent.com/microsoft/navcontainerhelper/dev/AppHandling/PsTestFunctions.ps1
and try it out?
I actually think that this might be a general problem, which might have caused strange issues when running tests too - there I scroll back in the same way now.

freddydk:
I think I was able to repro this and I think that other issues are caused by the same problem.
The latest fixes should work if you could try.
Thanks

freddydk:
Fix was shipped in 0.6.5.7, please reopen #944 if this doesn't solve the issue

PR #949, opened 2020-04-17
Support swarm services

tfenster:
When a container runs as Swarm service, then the volume will appear below Mounts, not below HostConfig.Binds


freddydk:
Thanks

Issue #959, opened 2020-04-28
"The property 'application' cannot be found on this object" error in Compile-AppInNavContainer


kine:
Hi,
have it this bug:

When I use the new property "applicationInsightsKey" in the app.json, the compilation ends with the error

````
The property 'application' cannot be found on this object ....
````
which originates on line 184 of the Compile-AppInNavContainer:

````
if (([bool]($appJsonObject.PSobject.Properties.name -match "application")) -and $appJsonObject.application)
    {
        $dependencies += @{"publisher" = "Microsoft"; "name" = "Application"; "version" = $appJsonObject.application }
    }
````

It is because the -match "application" match this line too... I recommend to use -eq instead -match if the property name should be exactly "application".

freddydk:
Shipped in 0.6.5.7

PR #960, opened 2020-04-28
Replace -match with -eq to solve #959

kine:
Replaced the -match which even partially match the property name to exact -eq comparison to prevent future bugs when new properties will be introduced which partially match existing names.


freddydk:
There are a lot of unwanted changes to this PR - space changes - could you remove these?

freddydk:
I did the -match to -eq in this and other scripts where applicable in another PR, thanks for the fix.

kine:
Ok, sorry for the format things, done automatically by VSCode formatting.

PR #975, opened 2020-05-05
Avoid error when used on machines without IE Engine (containers, etc).

ShadowXVII:
> ##[error]Invoke-WebRequest : The response content cannot be parsed because the Internet Explorer engine is not available, or 
> ##[error]Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again. 
>


freddydk:
Thanks

PR #1003, opened 2020-05-28
Fix time format in XUnit test result file

### LINKED ISSUE #1002: ##[warning]Atleast for one assembly start time was not obtained due to tag not present or parsing issue, total run duration will now be summation of time taken by each assembly
**Describe the issue**
In some pipeline runs I get this warning when publishing the test results:
##[warning]Atleast for one assembly start time was not obtained due to tag not present or parsing issue, total run duration will now be summation of time taken by each assembly

I've hunted it down to that this is depending on the time format of the container. I get this in the Finnish localization, and they have a dot (.) as time separator - not colon (:) as many others.
So in the test result file we get `run-time="11.25.21"` instead of `run-time="11:25:21"`, which is unexpected by DevOps...

Now I found out that the `datetime.ToString("HH:mm:ss")` does not do what you expect. The : in the format string says that it should use the current configured time separator, not : as we write. (ref: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#timeSeparator)

The solution is to wrap the : inside '
`datetime.ToString("HH':'mm':'ss")`

A PR to fix this is on it's way...

jwikman:
Fixes #1002


freddydk:
Thanks Johannes.

PR #1008, opened 2020-06-03
Update Get-NavContainerSharedFolders.ps1

svengrav:
A path is displayed twice if a path with backlash at the end is specified at container start. The optional removal of the backlsash ensures that the function returns the correct paths independently.
**Before:**
image
**After:**
image


freddydk:
Thanks, sorry for the delay

PR #1025, opened 2020-06-19
Get-BCArtifactUrl

waldo1001:
As promised - based on the API.

I didn't look into the bcinsider-part (yet).

### DIFF
--- a/Misc/Get-BCArtifactUrl.ps1
+++ b/Misc/Get-BCArtifactUrl.ps1
@@ -10,7 +10,9 @@
  .Parameter Version
   The version of Business Central (will search for entries where the version starts with this value of the parameter)
  .Parameter Select
-  All or only the latest (Default All)
+  All or only the latest (Default All):
+    - All: will return all possible urls in the selection
+    - Latest: will sort on version, and return latest version
  .Parameter StorageAccountName
   The StorageAccount that is being used where artifacts are stored (Usually should not be changed).
   .Example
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks
2 questions
What does it return if you specify All?
You have a limit=10 - is that on purpose? doesn't that mean that you will never get the US blob?

waldo1001:
Sorry - the limit=10 was a test that didn't work - it was still in my code.  I removed it.
All will return all possible urls within the current filter.  While latest will obviously return only the latest one...

freddydk:
for insider (and private artifacts store) support - I would rename the storageaccountname parameter to storageaccount - and if the storageaccount value doesn't have a . in it - add .azureedge.net - else keep as is (allows for embed partners to create a storage account with apps, pointing to MS platforms).

also - add a parameter called sasToken (default ""). If not "" - add that to the URL (before adding the other parameters - and add the other parameters with an & instead of ?) - this is the way insider sas tokens are shared on collaborate.

freddydk:
This is very cool:-)

waldo1001:
Should the sasToken from the document work?

freddydk:
The statement on SasToken above is wrong - it cannot just be added to the rest request.
I still think we want the full SasToken as a parameter, but the shared access key needs to be extracted and added in the request part.
I am happy to take this part if you don't have time?

waldo1001:
Thanks for documentation link.
I don't mind to give this a go ... .  I'll report back on monday!

waldo1001:
@freddydk , I included the sasToken.  Looking forward to your feedback

freddydk:
Did it work like that? I thought it wouldn't

waldo1001:
It did - I needed to provide the resource type (&restype=container)

freddydk:
Cool - will test tonight

freddydk:
Thanks Waldo

waldo1001:
Very welcome :-).

freddydk:
Found a few issues, have fixed them myself, will checkin tomorrow

waldo1001:
Wow .. hardly recognizable :-). Nice!

PR #1069, opened 2020-07-06
Custom artifact cache path

hemisphera:
This allows to specify a custom path where to cache downloaded artifacts. C:\ is usually pretty stuffed on our servers.

Sorry for whitespace-noise, apparently my VSCode formats stuff a little different.


freddydk:
The fix I will apply will count both the programdata folder AND the bcartifacts cache + some other things - and it will be a way to configure the module.
so basically, after installing the module you set the path and that will be written into a config.json somewhere, which is used on startup - something like that.

freddydk:
the problem with the cache path is that it cannot be longer than bcartifacts.cache - then you will fail elsewhere.

hemisphera:
OK, but at least it gives us the possibility to move it to another drive. Our build servers have a limited C:\ and most of it's data is stored on a second drive. This would help in the meantime.

freddydk:
In the next version, you can place a file in c:\ProgramData\NavContainerHelper with the following:

{
    "bcartifactsCacheFolder":  "d:\\bcartifacts.cache"
}

then it will use that as the cache folder.

At this time, that is the only setting and it needs to be set when importing the module (on PS start)
This will eventually be extended to also move the containerhelper folder and do other global settings - but this is a start for now.

freddydk:
new version should be out tonight sometime.

hemisphera:
Hmm. I'm curious to see the implementation details of this. Is it committed yet? 

I'd hate for this to become a machine-wide setting that I can not change on a per-session basis. I hope we can still override it somehow? Like ... that JSON would be better off being read into a global variable that gets exported from the module. This way we can still change it, without interfering with the JSON file itself.

freddydk:
That is a good point - Will make all global settings are overridable from code as well

Hent Outlook til iOS<https://aka.ms/o0ukef>
Fra: The Dumb <notifications@github.com>
Sendt: Monday, July 6, 2020 2:22:36 PM
Til: microsoft/navcontainerhelper <navcontainerhelper@noreply.github.com>
Cc: Freddy Kristiansen <Freddy.Kristiansen@microsoft.com>; State change <state_change@noreply.github.com>
Emne: Re: [microsoft/navcontainerhelper] Custom artifact cache path (#1069)


Hmm. I'm curious to see the implementation details of this. Is it committed yet?

I'd hate for this to become a machine-wide setting that I can not change on a per-session basis. I hope we can still override it somehow? Like ... that JSON would be better off being read into a global variable that gets exported from the module. This way we can still change it, without interfering with the JSON file itself.

freddydk:
ok - implementation is now
Add support for configuration file C:\ProgramData\NavContainerHelper\NavContainerHelper.config.json to allow for global configurations (bcartifactsCacheFolder for now)
Add support for configuration hashtable $navContainerHelperConfig - with runtime settings defaulted from code and from config file

c:\bcartifacts.cache is default
the navcontainerhelper.config.json overrides the default (if it exists)
$navcontainerhelperconfig.bcartifactscachefolder can override that in code

PR #1070, opened 2020-07-06
Fixing capitalization error in the vairable name

gierszewski:
Very minor fix in the parameter name (imagename --> imageName)


freddydk:
:-)

PR #1083, opened 2020-07-08
Limit SQL Server memory to half of the containers memory limit

StefanMaron:
While checking some memory issues with my container I discovered that the SQL Server does not have any memory limit. This can lead to problems where the SQL Server will consume all memory from the container.

If you pass a memory limit for the container, SQL Server memory limit will be set to half of the limit of the container. Up to a maximum of 10G. If no container memory limit is passed, SQL Server will not be limited either. 

PS: I do not unterstand why there are changes without anything changed.

The relevant changes are line number 1134 and 1475 to 1503

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -1131,6 +1131,7 @@ function New-NavContainer {
     if ("$memoryLimit" -eq "") {
         if ($isolation -eq "hyperv") {
             $parameters += "--memory 4G"
+            $memoryLimit = "4G"
         }
     } else {
         $parameters += "--memory $memoryLimit"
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks a lot for this contribution.
I probably won't include it as is - I have learned that every time I change something fundamental, some pipeline somewhere in the world breaks.
I am considering how I can release things like this without adding parameters and without breaking people.

StefanMaron:
This is perfectly okay for me. I just think that an unlimited SQL Server can lead to problems and its best practice to set a limit. 

I wanted to provide a solution which works as far as i can tell. If you can reuse any of this I am okay if it wont be merged as it is 👍

freddydk:
This was included.
I learned afterwards though that SQL Express can maximum take 1Gb of memory, so this is really only for limiting SQL below that.

Issue #1088, opened 2020-07-08
Concurrency issues with Artifacts


jwikman:
**Describe the issue**
We've got a lot of pipelines running in parallell on the same (hosted) build server.
Earlier, when we had different images published for each locale and version, docker daemon handled concurrency when several pipelines requested to download the same image.

Today I've been running a lot of pipelines with the new Artifact approach, and I've seen some strange issues that seems to be related to concurrency.

I've only been able to reproduce one of them by now, I will add more if I'm able to repro later.

### Repro
1. Clear your bcartifacts cache to be sure that artifacts are downloaded
1. Also clear you docker image cache if you've got any images called 'myimage' :)
1. Open PowerShell ISE (as admin) and open two tabs (two separate PowerShell sessions) with a new editor window in each tab.

4. In the first tab you paste this:
```
$credential = New-Object pscredential 'super', (ConvertTo-SecureString -String "P@ssword" -AsPlainText -Force)
$parameters = @{
    "accept_eula"              = $true
    "accept_outdated"          = $true
    "containerName"            = 'test2'
    "artifactUrl"              = "https://bcartifacts.azureedge.net/sandbox/16.3.14085.14363/se"
    "imageName"                = 'myimage'
    "auth"                     = "NavUserPassword"
    "Credential"               = $credential
    "dns"                      = '8.8.8.8'
    "alwaysPull"               = $true
    "updateHosts"              = $true
    "includeAL"                = $true
    "includeCSide"             = $false
    "doNotExportObjectsToText" = $true
    "shortcuts"                = "none"
    "myScripts"                = @(@{'MainLoop.ps1' = 'while ($true) { start-sleep -seconds 10 }' })
    "doNotCheckHealth"         = $true
    "enableTaskScheduler"      = $false
}

New-NavContainer @parameters
```
5. In the second tab you paste this (The only differences is the containerName):
```
$credential = New-Object pscredential 'super', (ConvertTo-SecureString -String "P@ssword" -AsPlainText -Force)
$parameters = @{
    "accept_eula"              = $true
    "accept_outdated"          = $true
    "containerName"            = 'test1'
    "artifactUrl"              = "https://bcartifacts.azureedge.net/sandbox/16.3.14085.14363/se"
    "imageName"                = 'myimage'
    "auth"                     = "NavUserPassword"
    "Credential"               = $credential
    "dns"                      = '8.8.8.8'
    "alwaysPull"               = $true
    "updateHosts"              = $true
[...]

PR #1095, opened 2020-07-09
Mutex to prevent concurrency issues

jwikman:
This PR tries to solve some of the issues in #1088 

If several threads downloads artifacts and builds docker images at almost the same time, things goes much slower since they are fighting about resources as CPU, bandwidth, disk space and disk performance.

With a mutex to prevent different processes to do the same thing at the same time, we let the first process finish it's task before the next process checks if this is needed.

My tests shows significant time gain.

### Repro code

Clear bcartifacts.cache and run below in two PS sessions, but change containerName to be different
````
$credential = New-Object pscredential 'super', (ConvertTo-SecureString -String "P@ssword12" -AsPlainText -Force)
$parameters = @{
    "accept_eula"              = $true
    "accept_outdated"          = $true
    "containerName"            = 'test2'
    "artifactUrl"              = "https://bcartifacts.azureedge.net/sandbox/16.3.14085.14363/gb"
    "imageName"                = 'myimage'
    "auth"                     = "NavUserPassword"
    "Credential"               = $credential
    "dns"                      = '8.8.8.8'
    "alwaysPull"               = $true
    "updateHosts"              = $true
    "includeAL"                = $true
    "includeCSide"             = $false
    "doNotExportObjectsToText" = $true
    "shortcuts"                = "none"
    "myScripts"                = @(@{'MainLoop.ps1' = 'while ($true) { start-sleep -seconds 10 }' })
    "doNotCheckHealth"         = $true
    "enableTaskScheduler"      = $false
}

Write-Host "It took $((Measure-Command -Expression { New-NavContainer @parameters }).TotalSeconds) seconds"
```` 

Clear bcartifacts.cache and run below in two PS sessions at the same time
```` 
Measure-Command -Expression {
Download-Artifacts -artifactUrl https://bcartifacts.azureedge.net/sandbox/16.3.14085.14363/us
}
```` 

### Without this change
Download the same artifacts in two processes: 105 seconds
Create new containers with in two processes, same imagename and artifactUrl: 1007s

### After this change
Download the same artifacts in two processes: 55 seconds
Create new containers with in two processes, same imagename and artifactUrl: 863s

ps. There seems to be some whitespace/line break changes as well, it was something VSCode did without showing the diffs. Sorry about that.

### CODE REVIEW

**Comment by jwikman** [comment]
I just realized that it might be needed an additional mutex for the platform, will update with that.

**Comment by jwikman** [comment]
Changed target branch to dev, you might want to prettify stuff if you decide to merge this.
Added a new mutex for downloading platform

**Comment by freddydk** [comment]
I haven't worked with mutex - what happens if something goes wrong and the thread holding the mutex crashes?

**Comment by jwikman** [comment]
> I haven't worked with mutex - what happens if something goes wrong and the thread holding the mutex crashes?

Good question.
I hadn't tested this, I took for granted that it worked as I expected it to (\*embarassed\*)... nowadays an exception is thrown, it wasn't like that once upon a time...

Killing the process having the mutex results in below error in the waiting thread:
`Download-Artifacts : Exception calling "WaitOne" with "0" argument(s): "The wait completed due to an abandoned mutex."`



I'll update to catch and handle the AbandonedMutexException.

**Comment by jwikman** [comment]
@freddydk Now I've updated with AbandonedMutexException handling. 

In a lot of cases AbandonedMutexException needs to be handled, because something that might be critical just crashed. But in our scenario we know that we will recover that kind of crash by just continuing.

You can easily test out the behavior by running below in several PS sessions and then kill the one that has the mutex (i.e. the one that are downloading) and see what happens with the waiting threads. Clear cache first. :)

```
Measure-Command -Expression {
Download-Artifacts -artifactUrl https://bcartifacts.azureedge.net/sandbox/16.3.14085.14363/us
}

```

Issue #1102, opened 2020-07-12
Remove-NavContainer does not remove folder with shortcuts from desktop


git-no:
Remove-NavContainer does not remove the shortcut Folder, neither shortcuts within the folder from desktop.

The folder with shortcuts was previously created with parameter `-shortcuts DesktopFolder` of command `New-NavContainer`.

freddydk:
tried to create a container and remove it again - works fine.
what name did you use for the container?
Did it leave one or all shortcuts in the folder?

freddydk:
Shipped in 0.7.0.14

PR #1106, opened 2020-07-13
fixed variable reuse at shortcut deletion

### LINKED ISSUE #1105: Shortcuts are not deleted when containers are deleted
Wenn removing a container the startmenu shortcuts wont be deleted.

source of the error is probably that the $name variable is edited several times (once for Desktop and once for CommonDesktop) --> the corresponding shortcut is no longer found in the second block of the function

https://github.com/microsoft/navcontainerhelper/blob/3f9383d77a9c3290617f7b36ab055df6be8cd6c4/Misc/Remove-DesktopShortcut.ps1#L23

ChrisBlankDe:
fixes #1105
#1102 could not reproduced with or without this fix

### DIFF
--- a/Misc/Remove-DesktopShortcut.ps1
+++ b/Misc/Remove-DesktopShortcut.ps1
@@ -20,9 +20,9 @@ function Remove-DesktopShortcut {
             }
             else {
                 $folderName = $Name.Split(' ')[0]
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks

PR #1113, opened 2020-07-14
Use 7zip if exist to expand archive

left-code:
For better performance use of 7zip instead of Expand-Archive when it's possible - on Azure and GitHub hosted Agents 7zip is installed by default.

I tested it to expand platform artifact and get following result:

- Using Expand-Archive - 57 seconds
- Using Expand-7zipArchive - 8 seconds

In case that 7zip tool is not found - system will use standard Expand-Archive function.

### DIFF
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -173,7 +173,7 @@ function Compile-AppInNavContainer {
         if (!(Test-Path "c:\build" -PathType Container)) {
             $tempZip = Join-Path $env:TEMP "alc.zip"
             Copy-item -Path (Get-Item -Path "c:\run\*.vsix").FullName -Destination $tempZip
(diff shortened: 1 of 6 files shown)


freddydk:
Very nice.
A few of the occurrences you have replaced with Expand-7zipArchive are inside a Invoke-ScriptInNavContainer, which means that the code is running inside the container.
The Expand-7zipArchive is in HelperFunctions in NavContainerHelper and as such cannot be reached from in there.
I found the one in Compile-AppInNavContainer and Convert-Txt2Al - not sure whether any of the others also are.

left-code:
Ah, you are right, not make sense to use it inside container (7zip is not there anyway).
I will check and update PR with only relevant replacement.

freddydk:
Correct 7zip is not and the helperfunctions isn't either.
Thanks

freddydk:
Thanks - will make some tests

Issue #1219, opened 2020-08-17
Generate-SymbolsInNavContainer.ps1 does not have BC alias


epernst:
Running Import-TestToolkitToBCContainer in a v14 container with Symbol generation enabled fails with:

_The term 'Generate-SymbolsInBcContainer' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
            at Import-TestToolkitToBcContainer, C:\Program Files\WindowsPowerShell\Modules\BcContainerHelper\1.0.2\ObjectHandling\Import-TestToolkitToNavContainer.ps1: line 238_

And it seems that its because there are no alias for Generate-SymbolsInBcContainer.

freddydk:
I will fix this at the call level - the missing alias is on purpose.
You can not generate symbols in a bc (AL) container - only in a nav (C/AL) container

freddydk:
Fix will be in a pre-release in a few hours.

freddydk:
Included in 1.0.3 as well.

PR #1220, opened 2020-08-17
Generate-SymbolsInNavContainer.ps1 does not have BC alias

epernst:
Generate-SymbolsInNavContainer.ps1 does not have BC alias

Ref: https://github.com/microsoft/navcontainerhelper/issues/1219


freddydk:
fix in a different way

PR #1294, opened 2020-09-18
Add multitenant support for unpublish

gunnargestsson:
We need to add the -tenant parameter if we have a multitenant container running and the app has been installed into the tenant scope.

### DIFF
--- a/AppHandling/UnPublish-NavContainerApp.ps1
+++ b/AppHandling/UnPublish-NavContainerApp.ps1
@@ -59,6 +59,10 @@ function UnPublish-BcContainerApp {
         if ($version) {
             $params += @{ 'Version' = $version }
         }
+        if ((Get-NAVAppInfo -ServerInstance $ServerInstance -Name $appName).Scope -eq "Tenant") {
+            $params += @{ 'Tenant' = $tenant }
+        }
+
         Write-Host "Unpublishing $appName"
         Unpublish-NavApp -ServerInstance $ServerInstance -Name $appName @params
     } -ArgumentList $appName, $unInstall, $tenant, $publisher, $version, $doNotSaveData, $force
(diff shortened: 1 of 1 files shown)


freddydk:
Is the Scope on Get-NavAppInfo available on NAV 2018 as well?
The same code is running and need to be resilient to old versions and not crash

gunnargestsson:
It is not but this code will not cause an error.

PS C:\Windows\system32> (Get-NAVAppInfo -ServerInstance DynamicsNAV110 -Name "Microsoft Pay Payments").Scope -eq "Tenant"
False

freddydk:
With Set-StrictMode -versioin 2.0 and $errorActionpreference = Stop - I think it will cause an error

gunnargestsson:
You are correct.

PS C:\Windows\system32> Set-StrictMode -version 2.0; $errorActionpreference = "Stop"
PS C:\Windows\system32> (Get-NAVAppInfo -ServerInstance DynamicsNAV110 -Name "Microsoft Pay Payments").Scope -eq "Tenant"
The property 'Scope' cannot be found on this object. Verify that the property exists.
At line:1 char:1
+ (Get-NAVAppInfo -ServerInstance DynamicsNAV110 -Name "Microsoft Pay P ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : PropertyNotFoundStrict

freddydk:
ContainerHelper runs with strict mode 2 and error stop.
I have a lot of places, where I check for properties before reading them - if you can do the same, I will take the PR

freddydk:
Needs to be -eq and not -match

gunnargestsson:
$appInfo.PSObject.Properties.Name is an array of all the property names.  That is why I use -match

freddydk:
-eq will do the same, but -match 'Scope' will also return true if 'ScopeEx' exists
Right now - that would not be a problem - but I have changed all occurances of -match to -eq all over because some property was added and something failed which I spend a lot of time fixing.

Issue #1299, opened 2020-09-22
Downloading Symbols that are already in the .alPackages folder.


Spycoclown:
PLEASE DO NOT INCLUDE ANY PASSWORDS OR TOKENS IN YOUR ISSUE!!!

**Describe the issue**
i get an issue that it cant not download the app file. (Symbols)
It don't need to download it because it is already there.

**Scripts used to create container and cause the issue**
```
 Compile-AppInBCContainer -containerName $containerName -credential $adminCredentials -appProjectFolder $appProjectFolder -appSymbolsFolder $buildSymbolsFolder -appOutputFolder (Join-Path $buildArtifactFolder $_) -UpdateSymbols:$false -AzureDevOps:$true -CopyAppToSymbolsFolder
```

**Full output of scripts**
```
C:\agent-01\_work\91\a\app\App1-System_1.0.0.1399.app successfully created in 76 seconds
App1-System_1.0.0.1399.app copied to C:\agent-01\_work\91\s\App1-System\.alPackages
C:\agent-01\_work\91\a\app\App1-System_1.0.0.1399.app
Using Version 1.0.0.1399
Compiling folder app test
Using Symbols Folder: C:\agent-01\_work\91\s\App1-System\.alPackages
Disabling SSL Verification
Downloading symbols: App1-System_1.0.0.0.app <-- why???
Url : https://123.22.153.189:7049/BC/dev/packages?publisher=GWS%20mbH&appName=App1-System&versionText=1.0.0.0&tenant=default
##[error]Compile-AppInBCContainer : Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an 
error: (404) Not Found."
At C:\agent-01\_work\91\s\Pipeline\Build-App\Compile-App.ps1:53 char:5
+     Compile-AppInBCContainer -containerName $containerName -credentia ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Compile-AppInBcContainer], MethodInvocationException
    + FullyQualifiedErrorId : WebException,Compile-AppInBcContainer

```

freddydk:
-updatesymbols will always try to download symbols - try to remove that flag.

Spycoclown:
Sry my mistake it is $false not true.
```
[...]

PR #1301, opened 2020-09-23
check app file is sync to containerSymbolsFolder

Spycoclown:
It helps in my sync problem, I'm not sure if it is global usable.

Issue: #1299


freddydk:
if you move the block one line up (inside the if statement), meaning that you only wait for it if the file was copied.
Then I will take it.

Thanks

freddydk:
I have copy/pasted your code and included it myself because another partner ran into the same issue this evening.
Thanks for your contribution.

PR #1334, opened 2020-10-05
Add ability to skip creating the image if it already exists

tfenster:
This adds a switch `-skipIfImageAlreadyExists` to `New-BcImage`. The use case is that I always want to call `New-BcImage` when running a container, but if the image already exists, I don't want to recreate it

### DIFF
--- a/ContainerHandling/New-NavImage.ps1
+++ b/ContainerHandling/New-NavImage.ps1
@@ -40,7 +40,8 @@ function New-BcImage {
         [switch] $includeTestToolkit,
         [switch] $includeTestLibrariesOnly,
         [switch] $includeTestFrameworkOnly,
-        [switch] $includePerformanceToolkit
+        [switch] $includePerformanceToolkit,
+        [switch] $skipIfImageAlreadyExists
 
     )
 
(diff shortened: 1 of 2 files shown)


ContainerHandling/New-NavImage.ps1
```diff
@@ -37,11 +37,12 @@ function New-BcImage {
...
-        [switch] $includeTestToolkit,
-        [switch] $includeTestLibrariesOnly,
-        [switch] $includeTestFrameworkOnly,
-        [switch] $includePerformanceToolkit
-
+        [switch] $includeTestToolkit,
+        [switch] $includeTestLibrariesOnly,
+        [switch] $includeTestFrameworkOnly,
+        [switch] $includePerformanceToolkit,
+        [switch] $skipIfImageAlreadyExists
```
tfenster, line 44: This is an actual change, it adds the new param

ContainerHandling/New-NavImage.ps1
```diff
@@ -104,300 +105,306 @@ function New-BcImage {
...
+        if ($multitenant) {
+            $imageName += "-mt"
+        }
     }
-    else {
-        $baseImageExists = docker images --format "{{.Repository}}:{{.Tag}}" | Where-Object { $_ -eq "$baseImage" }
-        if (!($baseImageExists)) {
-            Write-Host "Pulling non-existing base image $baseImage"
+
+    $exists = (DockerDo -command images -imageName $imageName | Out-Null)
```
tfenster, line 119: This is an actual change. It checks if an image with the name that would be used for the build already exists. If it exists and the new switch is yes, then it skips the rest

tfenster:
seems there where changes while I created this. I'll try again

PR #1335, opened 2020-10-05
add option to skip image create if the image already exists

tfenster:
this adds a switch param to New-BcImage which allows to suppress image creation if the image already exists. The use case is that I am managing container hosts and just want to send them the command "New-BcImage ...". If the image doesn't exist, it is created. If it already exists, it returns.

I don't know why but GitHub identifies a ton of changes in the New-NavImage.ps1 file. I even tried editing in the browser to avoid any automatic format or change, but it still has way more than I actually changed. I'll mark the actual changes with comments

### DIFF
--- a/ContainerHandling/New-NavImage.ps1
+++ b/ContainerHandling/New-NavImage.ps1
@@ -40,7 +40,8 @@ function New-BcImage {
         [switch] $includeTestToolkit,
         [switch] $includeTestLibrariesOnly,
         [switch] $includeTestFrameworkOnly,
-        [switch] $includePerformanceToolkit
+        [switch] $includePerformanceToolkit,
+        [switch] $skipIfImageAlreadyExists
 
     )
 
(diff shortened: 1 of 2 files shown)


ContainerHandling/New-NavImage.ps1
```diff
@@ -37,11 +37,12 @@ function New-BcImage {
...
-        [switch] $includeTestToolkit,
-        [switch] $includeTestLibrariesOnly,
-        [switch] $includeTestFrameworkOnly,
-        [switch] $includePerformanceToolkit
-
+        [switch] $includeTestToolkit,
+        [switch] $includeTestLibrariesOnly,
+        [switch] $includeTestFrameworkOnly,
+        [switch] $includePerformanceToolkit,
+        [switch] $skipIfImageAlreadyExists
```
tfenster, line 44: change

ContainerHandling/New-NavImage.ps1
```diff
@@ -104,300 +105,306 @@ function New-BcImage {
...
+        if ($skipDatabase) {
+            $imageName += "-nodb"
+        }
+        if ($multitenant) {
+            $imageName += "-mt"
+        }
+    }
     
-    $imageName
+    $exists = (DockerDo -command images -imageName $imageName | Out-Null)
```
tfenster, line 119: begin change

ContainerHandling/New-NavImage.ps1
```diff
@@ -104,300 +105,306 @@ function New-BcImage {
...
+    if ($exists -and $skipIfImageAlreadyExists) {
+        Write-Host "Skip building the image as it already exists"
+        $imageName
+    } else {
```
tfenster, line 123: end change

freddydk:
Not sure why, but there are a ton of changes when reviewing this - it is hard to see what actually changed and what not???

tfenster:
yes, as I wrote I ended up changing it directly in the browser as I thought that maybe VS Code was doing some formatting even if I used "save without formatting", but Github still shows lots of changes. Maybe it's because of this?

image

tfenster:
Oh yes, this seems to be the case. If you go to the "changed files" view, select the gear icon and enable "hide whitespace changes", then you see only the actual changes

image

freddydk:
Will have a look.
New-BcContainer already does this in a slightly different way. Maybe I should move that to New-BcImage and let New-BcContainer call over with the parameter.

freddydk:
I have just checked in a new New-BcContainer and a new New-BcImage which introduces this parameter.
It will force the rebuild if the image doesn't exist or if it was build on an old generic version or for a different OS version.
All checks from New-BcContainer has moved to New-BcImage.

tfenster:
Thanks a lot, that is even better! Do you already have an idea when you plan to release the next bccontainerhelper version?

freddydk:
If all tests passes it will auto-release as a preview version in ~2 hours.

PR #1358, opened 2020-10-12
Remove check for clientservicesport

gunnargestsson:
Since Client Services are not mapped via Traefik we should be able to specify the port we want to use.


freddydk:
Will allow this one.
If we get traefik v2 we might put this back in as clientservices will go through traefik v2

PR #1438, opened 2020-11-09
Check Traefik image before pulling

itsme112358:
Since Docker is starting to enforce pull limitations it makes sense to check if the image is already up to date instead of pulling every time.


freddydk:
I was actually planning to change this code to build the traefik image based on a base image, which matches the host, instead of downloading an old traefik image.

freddydk:
Taking this until I get the other one created, thanks

PR #1447, opened 2020-11-12
support alc features in Compile-AppInBcContainer

IceOnly:
We need the translation file in our build pipeline to check for missing translations.
I added all known features to Compile-AppInBcContainer CmdLet.


freddydk:
Could you add a line to Releasenotes?
Thanks

freddydk:
You would need to add a new headline 

1.0.14
support features...

PR #1463, opened 2020-11-18
Bug Fix: Param allowAppDatabaseWrite added to script block

svengrav:
The parameter **$allowAppDatabaseWrite** must be passed as param to the script block. Otherwise, the variable in allowAppDatabaseWrite script block will be always $false. Same behavior as with the switch doNotCopyDatabase.


svengrav:
image

freddydk:
Thanks, rookie mistake:-(

svengrav:
Glad to help, good work with the ContainerHelper, it really helps us a lot! 👌

PR #1570, opened 2020-12-24
Nav container helper

TGosselink:
Unfortunately the new parameter: -digestAlgorithm doesn't work correctly

Used Parameters:
Sign-BcContainerApp -appFile $_.FullName -pfxFile $MyPfx -pfxPassword $MyPassword -containerName $mycontainerName -digestAlgorithm sha256

Error returned:
SignTool Error: The /t option is incompatible with the /td option.

Reason:
/t should not be used with SHA256 but /tr instead.
In your code, that's an easy fix.

Source: https://www.thegeekstuff.com/2017/01/signtool-examples/

Example 2: Code Sign using SHA256 Algorithm
To code sign using SHA256, in the Windows command prompt, enter the following command.

signtool.exe sign /a /tr http://timestamp.geotrust.com/tsa /td sha256 /fd sha256 /v "c:\thegeekstuff.exe"
In the above command:

/tr – “tr” here stands for time stamp server RFC 3161. For this, you have to make sure the time stamp server is a RFC 3161 support. If not, this will return an error message as shown below. I’ve used the geotrust.com URL. You can also use this RFC 3161 URL: http://timestamp.digicert.com
/td – “td” here stands time server digest algorithm. This is used to request the specified digest algorithm (in this case, sha256) from the specified RFC 3161 time server. Make sure the /td switch is declared after the /tr switch. If you specify this before the /tr switch, then the timestamp will be returned from SHA1 algorithm and from SHA256 as you would hope for.
/fd – “fd” here stands for File Digest Algorithm. By default when you don’t specify this parameter, it will use the SHA-1 algorithm. In this example, it will use the SHA256 algorithm to digitally sign the file.
Error 1: If you specify /t option (instead of /tr) when you give /td, it will throw the following incompatible option error message.

C:> signtool.exe sign /a /t http://timestamp.verisign.com/scripts/timstamp.dll /td sha256 /fd sha256 /v "c:\thegeekstuff.exe"
SignTool Error: The /t option is incompatible with the /td option.


freddydk:
I think you are having a lot of old files in here

freddydk:
and it looks like the Sign-BcContainerApp is not here (maybe the fix I already checked in this morning is the same?)

PR #1700, opened 2021-02-12
Extending functions Export-ModifiedObjectsAsDeltas & Convert-Modified…

gintuoklis:
…ObjectsToAl with parameter -originalFolder

Adding additional parameter originalFolder to the functions  Export-ModifiedObjectsAsDeltas & Convert-ModifiedObjectsToAl.

Currently $originalFolder is looking at Extension folder based on container version. In scenarios when your container contains different application version from platform version it becomes a limitation. For e.g. our container is is running CU10, while base appication is CU5. Therefore, exposing  $originalFolder as parameter allows us to override it and specify our folder with origal files.


freddydk:
Thanks @gintuoklis

PR #1758, opened 2021-03-02
Fix bug with snap endpoint

adamLiveware:
When using Traefik the dev endpoint settings are overridden by the snap endpoint.
It causes the dev port to be inaccessible.


freddydk:
Looks like a typo from @tfenster 
Thanks for submitting

tfenster:
@freddydk oh yes, sorry :(
@adamLiveware thanks!

Issue #1736, opened 2021-02-25
Install all apps from a given folder.


jonaswre:
When passing a folder to an InstallApps Parameter in various scripts the CopyAppFilesToFolder throws an error in the line https://github.com/microsoft/navcontainerhelper/blob/0ae7815208bfac936c1b7e8d2ded50653055b2d7/HelperFunctions.ps1#L360.
 It would be very useful to be able to pass in folders and then have the CopyAppFilesToFolder iterate through the given folders and copy all apps within those folders.

freddydk:
You should just use
-installApps (get-childitem $myFolder -Filter '*.app' -Recurse)

much more flexible and gives you the ability to do recurse, filters and exclude special things.

jonaswre:
Thanks for the quick response. I've been doing that, but i would like to change as little as possible in your scripts from the Hello-World AL Project. This makes it easier to integrate changes.
Any reason not to do that in the helper itself?

freddydk:
I will consider checking whether the path is a folder and then grabbing all .app files recursively.

jonaswre:
Thanks that would be great.

freddydk:
Shipped in 2.0.6-preview355

jonaswre:
Thanks

PR #1769, opened 2021-03-08
Dependency handling using azure feed. 

jonaswre:
The dependency tree is getting more and more complex. Because of that fact a mechanism is needed to resolve these dependencies. 
We found that using azure feed works farely well for this scenario. 
The only draw back with azure feed is the enforced versioning theme. Since the revision part of the version no needs to be cut off. But for our product development this is acceptable. 

Our current workflow is that we publish each app from the pipeline artifact as its own package in to azure feed. 
From the build Pipeline the Resolve-DependenciesFromAzureFeed script downloads all dependencies recursively into a single folder. That folder can be used in the installApps paramenter since the enhancement in #1736.

Ultimately it would be great to include that functionality to the Run-AL-Pipeline script. But for now have that script not copied in very pipeline is enough.


ArtifactHandling/Publish-BuildOutputToAzureFeed.ps1
```diff
@@ -0,0 +1,79 @@
...
+            az artifacts universal publish `
+                --organization $organization `
+                --feed $feed `
+                --name $appJson.id `
+                --version $appVersion `
+                --description $appJson.name `
+                --path (Join-Path $tempAppOutFolder '.')   
+        }
+        finally {
+            Remove-Item -Path $tempAppFolder -Recurse -Force
```
jonaswre, line 75: @freddydk If I understand correctly Remove-Item should be in the finally so that what ever happens above the temp folder will be always cleaned up

freddydk: yes, maybe add an -erroraction SilentlyContinue to the remove-item, so that if a failure occurs and the removal fails, then the original error is shown and not the remove-item failure.

ArtifactHandling/Resolve-DependenciesFromAzureFeed.ps1
```diff
@@ -0,0 +1,125 @@
...
+ .Synopsis
+  Function for resolving dependencies using an AzureFeed
+ .Description
+  Function for resolving dependencies using an AzureFeed
+  To being able to resolve dependencies an artifact must be in the following format.
+  The app id must be in the artifact name so apps will be identified correctly.
+  The artifact can contain an .runtime.app file. If no app.json or .app file is present the .runtime.app fill be treated as a leaf in the tree.
+  If an app.json is present dependencies will be take from that file.
+  If no app.json but an .app file is present that will be extracted and used to find dependencies. 
+#>
```
freddydk, line 11: Could you add parameter descriptions.
And then please rename $appsPath to $appsFolder (and $outputFolder instead of $outPath)

ArtifactHandling/Get-AzureFeedWildcardVersion.ps1
```diff
@@ -0,0 +1,27 @@
+function Get-AzureFeedWildcardVersion {
```
freddydk, line 1: Synopsis, description and parameters

ArtifactHandling/Install-AzDevops.ps1
```diff
@@ -0,0 +1,24 @@
+function Install-AzDevops {
```
freddydk, line 1: Synopsis, description and parameters

freddydk:
Maybe we should rename the folder from ArtifactHandling to AzureFeedHandling (to avoid future confusion with BC artifacts)

jonaswre:
> AzureFeedHandling

Maybe PackageHandling? Since its kind of a package manager for bc

freddydk:
will be in next preview - probably 4-5 hours from now.

jonaswre:
@freddydk thanks

freddydk:
I am the one saying Thanks:-)

PR #1774, opened 2021-03-08
Strict mode got me bad

jonaswre:
In strict mode things work a bit diffrent. Specially with catches an $_ one needs to take care. 

Added an strict mode save check for properties in the app.json. 
Added an id variable so that the id could be printed without an error in the catch phrase. 

sorry for the trouble...


freddydk:
Is this a public feed?
Resolve-DependenciesFromAzureFeed -organization "https://dev.azure.com/TEST/" -feed "BCApps" -appsFolder $PSScriptRoot

jonaswre:
No its just a placeholder so that it will parse the app.json's. The test is not used to test the uploading itself. 

But if you're intrested we could extend your helloworld (create a new one) with a more complex dependency tree that uses Resolve-DependenciesFromAzureFeed function.

freddydk:
Ahh OK, thanks

Issue #1800, opened 2021-03-18
Restore-BcDatabaseFromArtifacts.ps1 needs SQL installed


kine:
Was trying the trick from Restore-BcDatabaseFromArtifacts to use admin module from artifact (outside container usage of the management module) but was not able because the libraries were not found. Module sqlps is part of SQL installation but I do not want to install that on my host. I was digging into it and have found that the libraries are part of sqlserver PS module. I slightly modified the code to load the libraries from there (preparing PR):
```
            import-module SqlServer
            $SqlModule = get-module SqlServer
            $Path = Split-Path $SqlModule.Path
            $Smo = [Reflection.Assembly]::LoadFile((Join-Path $Path 'Microsoft.SqlServer.Smo.dll'))
            $SmoExtended = [Reflection.Assembly]::LoadFile((Join-Path $Path 'Microsoft.SqlServer.SmoExtended.dll'))
            $ConnectionInfo = [Reflection.Assembly]::LoadFile((Join-Path $Path 'Microsoft.SqlServer.ConnectionInfo.dll'))
            $SqlEnum = [Reflection.Assembly]::LoadFile((Join-Path $Path 'Microsoft.SqlServer.SqlEnum.dll'))
                    
            $OnAssemblyResolve = [System.ResolveEventHandler] {
                param($sender, $e)
                if ($e.Name -like "Microsoft.SqlServer.Smo, Version=*, Culture=neutral, PublicKeyToken=89845dcd8080cc91") { return $Smo }
                if ($e.Name -like "Microsoft.SqlServer.SmoExtended, Version=*, Culture=neutral, PublicKeyToken=89845dcd8080cc91") { return $SmoExtended }
                if ($e.Name -like "Microsoft.SqlServer.ConnectionInfo, Version=*, Culture=neutral, PublicKeyToken=89845dcd8080cc91") { return $ConnectionInfo }
                if ($e.Name -like "Microsoft.SqlServer.SqlEnum, Version=*, Culture=neutral, PublicKeyToken=89845dcd8080cc91") { return $SqlEnum }
                foreach($a in [System.AppDomain]::CurrentDomain.GetAssemblies()) {
                    if ($a.FullName -eq $e.Name) { return $a }
                }
                return $null
            }
            [System.AppDomain]::CurrentDomain.add_AssemblyResolve($OnAssemblyResolve)
```

[...]

PR #1801, opened 2021-03-18
Using SqlServer PS module to load the Smo libraries

kine:
Solve #1800

### DIFF
--- a/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
+++ b/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
@@ -84,8 +84,12 @@ function Restore-BcDatabaseFromArtifacts {
         if ($multitenant) {
             $dbName = "$($databasePrefix)tenant"
         }
-        Write-Host "Restoring database $dbName"
-        New-NAVDatabase -DatabaseServer $databaseServer -DatabaseInstance $databaseInstance -DatabaseName $dbName -FilePath $bakFile -DestinationPath (Join-Path $smoServer.RootDirectory "DATA\$($databaseprefix -replace '[^a-zA-Z0-9]', '')") | Out-Null
+        $DefaultDataPath = (Invoke-SqlCmd `
+            -ServerInstance $databaseServerInstance `
+            -Query "SELECT SERVERPROPERTY('InstanceDefaultDataPath') AS InstanceDefaultDataPath" ).InstanceDefaultDataPath
+            
+        Write-Host "Restoring database $dbName into $DefaultDataPath"
+        New-NAVDatabase -DatabaseServer $databaseServer -DatabaseInstance $databaseInstance -DatabaseName $dbName -FilePath $bakFile -DestinationPath (Join-Path $DefaultDataPath "$($databaseprefix -replace '[^a-zA-Z0-9]', '')") | Out-Null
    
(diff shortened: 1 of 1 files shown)


freddydk:
Very nice, have been thinking about this.

It does however break the code for those who has SQL installed and not SqlServer (like me)
If you could modify the code to do:

```
Import-Module sqlps -ErrorAction SilentlyContinue
$sqlModule = get-module sqlps
if ($sqlModule) {
    $smoServer = New-Object Microsoft.SqlServer.Management.Smo.Server $databaseServerInstance

    #load libraries the old way

}
else {
    Import-Module SqlServer -ErrorAction SilentlyContinue
    $sqlModule = get-module SqlServer
    if ($sqlModule) {
        
        # load libraries the new way

    }
    else {
        throw "You need to have a local installation of SQL or you need the SqlServer PowerShell module installed"
    }
}
```

kine:
Have added line to remve the resolver else the job ends with error.

freddydk:
It looks like smoServer is used in the codepath still and never set

freddydk:
smoserver is used in line 90
New-NAVDatabase -DatabaseServer $databaseServer -DatabaseInstance $databaseInstance -DatabaseName $dbName -FilePath $bakFile -DestinationPath (Join-Path $smoServer.RootDirectory "DATA\$($databaseprefix -replace '[^a-zA-Z0-9]', '')") | Out-Null

kine:
Ok, will check that, I have overlooked that. We can get the path in other way I think...

kine:
Which path it should be? Path where the database should be restored? It means default path for databases on the server?

freddydk:
If not specified (and restoring on own server) - it tries to use the path in the .bak file:-(
That is the reason for the construct.
Maybe it works with a relative path - I can test that, stay tuned

kine:
I have used t-sql to get the default data path from the SQL. Can you try with that?

kine:
I have stopped to use the New-NAVDatabase because it is not renaming the db files to new database names thus possible overwrite on the server when multiple databases are restored and the target will be same folder. Now I am using Restore-SqlDatabase with RelocateFile parameter.

freddydk:
should be this: 
-DestinationPath (Join-Path $DefaultDataPath ($databaseprefix -replace '[^a-zA-Z0-9]', ''))

but you need the import-modules before you use invoke-sqlcmd (on my machine it fails otherwise)

kine:
hmmm, no need for import-module on my. What it is missing on yours? 

I have different problem - it wants to restore to path where the prefix is subfolder but this doesn't exists on the server, thus restore fails (it needs to have this folder already existing).

kine:
Ok, fixed the import-module issue, now I am importing the modules before getting the default path.

Still there is problem when the target path doesn't exists, but I think it was same even in original version.

freddydk:
could  you remove the second 

```
if ($sqlmodule) {
```
and
```
                else {
                    throw "You need to have a local installation of SQL or you need the SqlServer PowerShell module installed"
                }
```

freddydk:
The reason is that sqlps is not in the modules folder and is not automatically imported upon usage.
The other modules gets automatically loaded when you use them the first time.

freddydk:
Thanks Kamil - great addition to the function.

Issue #1793, opened 2021-03-17
Setup-TraefikContainerForNavContainers | Use my own certificate | traefik.toml missing link to certificate


hvandenborn:
Freddy,

I setup Traefik with this "Setup-TraefikContainerForNavContainers" command, Then Started a container with traefik and that worked, Only. 
if i access the URL in our domain, you get the message that the HTTPS is not OK, hackers are stealing your passwords etc etc.
So i try to use our Wildcard certificate of our domain to make this work.
this server is only accessible on our internal domain, 

but if i call Setup-TraefikContainerForNavContainers:
Setup-TraefikContainerForBcContainers -PublicDnsName $PublicDNS -CrtFile $CrtFile -CrtKeyFile $CrtKeyFile [-Recreate]
it will copy the certificates to 
![image](https://user-images.githubusercontent.com/48794854/111456835-c0963880-8717-11eb-832f-ebb724e0ccaf.png)
And recreate the traefik container,
then i make a new BCcontainer 
![image](https://user-images.githubusercontent.com/48794854/111456939-ddcb0700-8717-11eb-93dc-eb6cce7614b3.png)
Still the certificate is not used,

now if i check the article 
https://www.axians-infoma.de/techblog/traefik-support-for-navcontainerhelper-the-nav-arm-templates-for-azure-vms-and-local-environments/
It points out that traefik.toml
needs to be eddited with:
![image](https://user-images.githubusercontent.com/48794854/111457417-73ff2d00-8718-11eb-9026-20850e67bcc4.png)
is this something the Setup-TraefikContainerForNavContainers forgets? because its not in my traefik.toml **It would be very nice if this part of the setup is also done, or if there is a switch or something to do this automaticly for using your own Certificate.** 

the only way i can make it work with my certificate is like https://www.axians-infoma.de/techblog/traefik-support-for-navcontainerhelper-the-nav-arm-templates-for-azure-vms-and-local-environments/ says,
change the certificate filenames from certificate to traefik
and add the TLS section and remove the [acme] section including [[acme.domains]]
then it works.

![image](https://user-images.githubusercontent.com/48794854/111469200-b3cd1100-8726-11eb-9eb0-6d864f58b9c6.png)

I also tried NOT to change te certificate filenames but changed the reference path in the TLS setion of traefik.toml from traefik.crt and .key to certificate.crt and .key, but that does not work. 

My traefik.toml after generation
```

------ domainname is replaced with MyDomain -------

------------------------------
[...]

PR #1809, opened 2021-03-19
Issue #1793

freddydk:


freddydk:
@tfenster could you review this?

tfenster (approved):
Looks good to me, but I have to admit that I've never used custom certs again since initially blogging about it because the Let's Encrypt stuff works so well and I never had a need to use anything else

freddydk:
Great, thanks

PR #1813, opened 2021-03-21
Allow arrays as path parameter for font-related functions

### LINKED ISSUE #1812: Idea: Allow array of paths for -addFontsFromPath
**Describe the idea**
I have two barcode fonts in my repo that I want to add to the container on New-BcContainer, but I also have to add the default fonts of windows (else my reports are barcode only). I could only add my windows fonts, but it could happen, that a developer does not have the barcode fonts installed on their host.

**Example usage**
```
New-BcContainer `
    ...
    -addFontsFromPath @("C:\windows\fonts\", ".\myfonts\") `
    ...
```

PS: I know I could just call "Add-FontsToBcContainer" after my New-BcContainer call.

Would you accept a PR for this?

dannoe:
Fixes #1812

Is the line ending normalization in ContainerHandling/New-NavImage.ps1 okay?

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -93,7 +93,7 @@
  .Parameter multitenant
   Setup container for multitenancy by adding this switch
  .Parameter addFontsFromPath
-  Enumerate all fonts from this path and install them in the container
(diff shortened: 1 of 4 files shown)


freddydk:
Looks very good, thanks

Issue #1867, opened 2021-04-06
Multiple traefik containers targeting same external database causing HasServerCrashed


sirhc101:
**Describe the issue**
We have setup two containers targeting the same external database. Both containers using traefik. Both containers are working pretty well if the other one is not running/started. As soon as we start the other container we encounter the following behavior:

- Session are timed out very fast. If you do not work for a few seconds you're kicked out 
- In event log we see this warning almost every 30 seconds:
![image](https://user-images.githubusercontent.com/39621236/113696320-213ee280-96d2-11eb-8bf6-a1c4c757f170.png)

I searched a lot and cannot find any reason. As I said when running just one container everything works well. Unfortunately we need both containers due to the fact AL language extension in VSCode is not supporting AAD auth. for other the `businesscentral.dynamics.com` sandboxes. So the second container is using `UserPassword` auth.

I've just one idea: In SQL Server Management Studio I've seen that both containers are using the same hostname. They both using the same server instance (BC). So I assume that the problem is caused by the "clean-up" or "health check" of server instances. In `[Server Instance]` table in SQL Server I see both server instances, but one does have the status "2"
![image](https://user-images.githubusercontent.com/39621236/113697188-28b2bb80-96d3-11eb-8785-55d9b0a29c27.png)

I saw in NavContainerHelper code that, when using traefik, the hostname is always the first part from the public dns name (subdomain), but is there are reason why the hostname of the computer needs to be this name? I tried `Rename-Computer` but of course it doesn't work because everytime the container restarts the hostname is set again.

Is there a way to change this or do you think I'm completely wrong and the problem is somewhere else?

Thank you in advance.

freddydk:
This could very well be due to the hostname.
You could try to modify the line setting the hostname to traefikhostname - to use the containername instead.
I am not entirely sure what impact this has and why the traefik hostname is set as it is.
Seems strange that all containers in a traefik environment will have the same hostname
Maybe @tfenster recalls the reason why?

sirhc101:
Update: I've added a new subdomain, so the containers do now get different hostnames and the problem seems to be gone. I'll try to modify the `New-NavContainer` code a bit to figure out how we can handle this a little better.

tfenster:
> This could very well be due to the hostname.
> You could try to modify the line setting the hostname to traefikhostname - to use the containername instead.
> I am not entirely sure what impact this has and why the traefik hostname is set as it is.
[...]

PR #1872, opened 2021-04-07
Running multiple traefik containers using same external database

sirhc101:
As described in issue #1867 when creating mutliple traefik containers using the same public dns name (e.g. `bc.domain.com`) all containers will get the first part of the public dns name as hostname (e.g. `bc`). This leads to HealthCheck issues and the server instances are battling against each other. The result of this is that a user get kicked out really fast (after a few seconds of inactivity). It also seems to make the complete environment extremely slow.

The problem can be solved by two ways:

1. Create subdomains for each container so every container will get a unique hostname
2. Remove the link in between public dns name and hostname

Option 1 is leading to another problem. With `Setup-TraefikContainerForBcContainers` you specify _**one**_ public dns name for which the acme challenge (*Lets Encrypt*) will be performed. If you now create a new subdomain with a different public dns name , for example `bc.domain.com` & `bc-test.domain.com`, you will not get a valid certificate for `bc-test.domain.com` and Browser, VSCode, Business Central App will complain about it.
For this I've added the command `Add-DomaintToTraefikConfig`. This command is included in `New-BcContainer` and checks the `traefik.toml` file for the acme challenge domains. If not already present as main domain or subdomain the public dns name will be added automatically.

> **Note**<br>As far as I noticed it is necessary to restart the traefik container after changing `traefik.toml` to retrieve the certificate. `Add-DomaintToTraefikConfig` is telling the user to do this after the container has been created manually. It's just `docker restart <traefik container name>`.

In #1867 @tfenster mentioned that the reason for syncing container hostname and public dns name was that "Web Client or/and NST" are not working correctly. I've tested the following scenarios after removing this:

- Web Client with Edge & Chrome **[OK]**
- Mobile Business Central App (Android) **[OK]**
- Windows Business Central App **[OK]**
- API Call using Postman **[OK]**

I cannot find any weird behavior after this. Anyway, as mentioned in #1867, I added configuration setting `TraefikUseDnsNameAsHostName` (default: false) to `bcContainerHelperConfig`. This is restoring the original behavior.

### DIFF
--- a/BcContainerHelper.psm1
+++ b/BcContainerHelper.psm1
@@ -219,6 +219,7 @@ Check-BcContainerHelperPermissions -Silent
 . (Join-Path $PSScriptRoot "ContainerHandling\Flush-ContainerHelperCache.ps1")
 . (Join-Path $PSScriptRoot "ContainerHandling\Get-LatestAlLanguageExtensionUrl.ps1")
 . (Join-Path $PSScriptRoot "ContainerHandling\Get-AlLanguageExtensionFromArtifacts.ps1")
+. (Join-Path $PSScriptRoot "ContainerHandling\traefik\Add-DomainToTraefikConfig.ps1")
 
 # Object Handling functions
 . (Join-Path $PSScriptRoot "ObjectHandling\Export-NavContainerObjects.ps1")
--- a/BcContainerHelper.psm1
+++ b/BcContainerHelper.psm1
@@ -69,6 +69,7 @@ function Get-ContainerHelperConfig {
                 "vn" = "w1"
                 "za" = "w1"
             }
+            "TraefikUseDnsNameAsHostName" = $false
         }
         $bcContainerHelperConfigFile = "C:\ProgramData\BcContainerHelper\BcContainerHelper.config.json"
         if (Test-Path $bcContainerHelperConfigFile) {
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -1,4 +1,4 @@
-﻿<# 
+<# 
  .Synopsis
   Create or refresh a NAV/BC Container
  .Description
@@ -1564,7 +1564,7 @@ if ($multitenant) {
         $dlRule="PathPrefixStrip:${dlPart}"
 
         $traefikHostname = $publicDnsName.Split(".")[0]
-        
+
         $traefikConfig = Get-Content
 
         $webPort = "443"
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -636,10 +636,6 @@ function New-BcContainer {
         if (-not $PublicDnsName) {
             throw "Using Traefik only makes sense if you allow external access, so you have to provide the public DNS name (param -PublicDnsName)"
         }
-        
-        Add-DomainToTraefikConfig `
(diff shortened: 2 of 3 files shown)


freddydk:
You should remove the bcContainerHelperPath parameter from the function and use the variable called $hostHelperFolder in the function instead - this is set to the containerhelper folder on the host in the .psm1 file.

freddydk:
I am not entirely sure I understand what the Add-DomainToTraefikConfig is doing?
Is that still needed when you change the hostname?

sirhc101:
> I am not entirely sure I understand what the Add-DomainToTraefikConfig is doing?
> Is that still needed when you change the hostname?

"Needed" is the wrong term I would say. ;)

To answer directly: No it's not needed.

But, just imagine you have a server somewhere and you run two containers there. Now you want to assign two different public dns names:
 - bc.yourDomain.com
 - bc-test.yourDomain.com

This is currently not possible. At least not possible w/o manually touching the `traefik.toml` file. The container will be created w/o any error, but as soon as you try to open the web client you're receiving a certificate warning. This is because traefik obtains only certificates for configured domains (at least in v1). So it would only optain a certificate for `bc.yourDomain.com` but not for `bc-test.yourDomain.com`.

`Add-DomainToTraefikConfig` is reading the `traefik.toml` file and searching for the `[[acme.domains]]` part which is relevant for http challenge (Lets Encrypt). Traefik does have to configuration nodes inside `[[acme.domains]]`. `main` specifies the "main domain" (or lets say first domain) and `sans` specifies additional or subdomains. 
`Add-DomainToTraefikConfig` is looking for the current public dns name in `[[acme.domains]]`. If it's already there everything is fine. Otherwise it will be added as `sans` (additional or subdomain).

The code is a little bit weird because I've not found any way to parse toml files. Thats why it reads the toml file line by line and works with regular expressions to identify the specific line.

By the way. In default toml file, which is created by BcContainerHelper there is no `sans` configuration because you assume there is only one public dns name. So `Add-DomainToTraefikConfig` is adding the `sans` configuration if needed.

At the end, based on my example above, the `traefik.toml` would look like this:
```toml
entryPoint = "https"
[acme.httpChallenge]
entryPoint = "http"
[[acme.domains]]
   main = "bc.yourDomain.com"
   sans = ["bc-test.yourDomain.com"]
```
`sans` can configured with multiple additional dns names, like:
```toml
[[acme.domains]]
   main = "bc.yourDomain.com"
   sans = ["bc-test.yourDomain.com","bc-dev.yourDomain.com","bc-stage.yourDomain.com"]
```

sirhc101:
> You should remove the bcContainerHelperPath parameter from the function and use the variable called $hostHelperFolder in the function instead - this is set to the containerhelper folder on the host in the .psm1 file.

Ok, I changed this with commit [0127eaf](https://github.com/microsoft/navcontainerhelper/pull/1872/commits/0127eaf55079a3278ceb6dd7b9c06decfd67064e)

freddydk:
Ok, got it.
I would prefer to remove the call from New-BcContainer to Add-DomainToTraefikConfig, but leave the function in.

Meaning that people can use the function before calling New-BcContainer and restart traefik.

sirhc101:
I removed it from `New-BcContainer` (from my perspective I would prefer a switch in `New-BcContainer` but it's ok for me. I know what to do ;)

freddydk:
Thanks

PR #1976, opened 2021-06-21
Added minimum version for generatecrossreferences

kfuglsang:
The new `generatecrossreferences ` flag is only supported by alc-compilers in BC18 and newer.
Older compilers will fail if the parameter is supplied.


freddydk:
Good idea, thanks

PR #2013, opened 2021-08-02
Convert-BcAppsToRuntimePackages: Allow to skip failing apps, run scriptblock after each processed runtime

MODUSCarstenScholling:
Convert-BcAppsToRuntimePackages is an easy to use bulk command without caring about container creation and file exchange.
Most times this is a better choice over Get-BcContainerAppRuntimePackage for simplicity.
However, there could be reasons to allow apps failing without breaking the whole process. And there is an option to handle processing for each created runtime separately.

 .Parameter skipFailingApps
  If set, a failing app (compiler error or anything else) does not stop the whole process but continues with the next app.
 .Parameter afterEachRuntimeCreation
  A script block to be executed after an app has been converted to a runtime package.
  The parameters are 'appFile', containing the source path and 'runtimeFile' containing the path
  to the newly extracted runtime file or $null if the process has failed and skipFailingApps has been set.


freddydk:
Will review this tomorrow, sorry for the delay

MODUSCarstenScholling:
Huh? You deleted the target branch?

freddydk:
That was a mistake - will recreate it.

freddydk:
Sorry

freddydk:
If you can recreate the PR?

freddydk:
either on master or dev

PR #2022, opened 2021-08-12
Support applicationInsightsConnectionString in app.json

TheNCuber:
Per runtime version 7.2 applicationInsightsKey is being deprecated by Microsoft (see link below). This change supports the replacement: **applicationInsightsConnectionString**

Microsoft Link: [https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-json-files](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-json-files)

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -1030,24 +1030,39 @@ Write-Host -ForegroundColor Yellow @'
     }
     catch {}
 
-    if ($app -and $applicationInsightsKey) {
-        if ($appJson.psobject.Properties.name -eq "applicationInsightskey") {
-            $appJson.applicationInsightsKey = $applicationInsightsKey
+    $bcVersion = [System.Version]$artifactUrl.Split('/')[4]
+    if($app) {
+        if(($appJson.runtime -ge 7.2) -or (!$appJson.runtime -and $bcVersion -ge [System.Version]"18.2")) {
+            if($applicationInsightsConnectionString) {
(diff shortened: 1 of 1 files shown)


freddydk:
Thanks
We should probably only set the connectionstring if runtime is 7.2 or higher - and only set the key if connectionstring is empty or runtime is less than 7.2.
This way you can have both values in a keyvault and specify both to all pipelines and the run-alpipeline will select the right value.
Make sense?

TheNCuber:
@freddydk Totally makes sense to me, I hesitated to implement this though, because according to the docs (same link as above) the runtime parameter is **not** mandatory. What do we do, if the user did not specifiy the runtime version in the app.json?

freddydk:
Good catch

This construct:
$bcVersion = [System.Version]$artifactUrl.Split('/')[4]
Gives you the version we are building against, then

If (runtime >=7.2 or (runtime is not set and $bcversion -ge [System.Version]"18.2.0.0") {
   # use connection string if set - else key
}
else {
   # use key
}

It becomes a little cumbersome - but it has the advantage that all pipelines will "just" work and the complexity is here instead of in all yaml files.

TheNCuber:
There you go: https://github.com/microsoft/navcontainerhelper/pull/2022/commits/8a3a7e2f872127e81e5950c176bb7918176f1a77. Maybe someone from the community is able to come up with a nice simplification that I overlooked, but I guess that piece of logic is needed here.

freddydk:
Thanks - will do a few tests and merge

PR #2024, opened 2021-08-17
Added parameter for a Custom Code Cop to Compile-AppInBcContainer

jfamvg:
This pull requests adds a new param to the Compile-AppInBcContainer so one or more Custom code cop's can be used when compiling apps.
This feature is following the same design as the Ruleset, the .dll file needs to be in the directory that is copied to the container in order to be used.
Also I tried to match the coding & documenting style.


freddydk:
Thanks, will run a few tests and merge

freddydk:
Thanks for the contribution

PR #2033, opened 2021-08-20
fix multitenant unpublish

lktraser:
When an app exists in the with the same version in multiple tenants it's impossible to unpublish with the bccontainerhelper.

The condition here doesn't work when a list is returned by Get-NAVAppInfo.

I'm not sure if there is a better way to test if 'scope' exists in the list but I also can't see a scenario where the property doesn't exist.

### DIFF
--- a/AppHandling/UnPublish-NavContainerApp.ps1
+++ b/AppHandling/UnPublish-NavContainerApp.ps1
@@ -79,7 +79,13 @@ try {
             $params += @{ 'Version' = $version }
         }
         $appInfo = (Get-NAVAppInfo -ServerInstance $ServerInstance -Name $name @params)
-        if ([bool]($appInfo.Scope -contains 'Tenant')) {
+        $tenantScope = $false
+        Get-NAVAppInfo -ServerInstance $ServerInstance -Name $name @params | ForEach-Object {
+            if ([bool]($_.PSObject.Properties.Name -eq 'Scope')) {
+                if ($_.Scope -eq "Tenant") { $tenantScope = $true }
+            }
+        }
+        if ($tenantScope) {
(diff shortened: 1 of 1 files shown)


freddydk:
In which case does Scope return a list?
Is Scope then an array? (what does $appInfo.Scope.GetType() return)
or is it a string?

lktraser:
Scope is never a list. But $appinfo can be (in case more than 1 tenant has the same app with same version) and then 
`if ([bool]($appInfo.PSObject.Properties.Name -eq 'Scope'))`
does not work as expected. Then then params are passed to the container without the tenant and BC can't identify from which tenant the app should be unpublished.

freddydk:
Sorry for the delay - I think the reason for 
if ([bool]($appInfo.PSObject.Properties.Name -eq 'Scope')) {

is that 'scope' isn't on appInfo in NAV 2018 (maybe even BC 14) and earlier versions.

This fix would cause these versions to fail on unpublishing then.

I guess this would be a better fix:

```
        $tenantScope = $false
        Get-NAVAppInfo -ServerInstance $ServerInstance -Name $name @params | ForEach-Object {
            if ([bool]($_.PSObject.Properties.Name -eq 'Scope')) {
                if ($_.Scope -eq "Tenant") { $tenantScope = $true }
            }
        }
        if ($tenantScope) {
            $params += @{ 'Tenant' = $tenant }
        }
```

lktraser:
Sounds good to me. I adopted that aproach for the PR.

freddydk:
You should remove this line as well:
$appInfo = (Get-NAVAppInfo -ServerInstance $ServerInstance -Name $name @params)

PR #2041, opened 2021-08-26
Allow runtime packages as previous apps run Run-AlPipeline

MODUSCarstenScholling:
For apps, there might be a requirement for runtime packages as dependent apps. These runtime packages could be different versions for the previous or current main app. So sorting must be possible even with these runtime packages.

In my opinion we cannot use install apps here, because the dependent versions are different. And as a small addition, the code is some seconds faster, omitting extraction.


MODUSCarstenScholling:
Any news here Freddy?

freddydk:
Looks good - but QQ - the previousApps are also used for AL Cops - apps specified in previousApps will be run through AppSourceCop - won't that be a problem?

MODUSCarstenScholling:
Hmm, that would "only" break my scenario but the changes would not have any general impact, right?

freddydk:
You are right - in that case things would fail beforehand - but the question is whether we should add a check to NOT run cops if runtime package?

MODUSCarstenScholling:
We have the process running for some time now. Here it is always dependencies being a runtime package, never the actual app. Not sure if someone would provide a runtime package as previous app of the old main app. If that would cause any issues in the future, I  would provide an update.

PR #2053, opened 2021-09-02
Fixing bug that downloaded all apps from first lvl

jonaswre:
There is a bug which doesn't allow you to have an test app with a dependency to the main app in the same folder (which it should).
In that case app will be resolved from feed as a dependency from test and installed. 
This will cause the build of app to fail, since it's already present in another version. 

I fixed this issue by including an option to ignore dependencies and always add the first lvl app ids to it. 

Besides fixing the issue this gives you the power, to for example ignore microsoft apps or anything that should not be resolved from feed.


freddydk:
Won't this change break anybody using the function already?

jonaswre:
Well I guess  it might. If somebody relies on this behavior, but I can't see a way how the current behavior is useful to anybody.

I could Include a switch which can be used to get the old behavior. Or the new behavior. But for me the old behavior is a bug.

freddydk:
Ok, thanks

PR #2061, opened 2021-09-03
BcSaaS scripts only works on MS environment

### LINKED ISSUE #2060: BcSaaS scripts only works on MS (businesscentral.dynamics.com) environment
**Describe the issue**
Base Url is hardcoded on BcSaaS related scripts so it's not possible to use the scripts with other environments other than the Microsoft's one.

A pull request will be open with suggested changes.

ricardopaiva:
Fixes #2060.


freddydk:
Looks good, the only change I would like to request is that with every place a URL is added as a parameter (like baseurl), there should be a line like this in the code before it is used:

$baseUrl = $baseUrl.TrimEnd('/')

This ensures that if people transfer a url with trailing /, it still works.

Thanks

ricardopaiva:
I see your point. I've updated the PR.
I've added that line in the beginning of each function, to be coherent.
Please check.

Issue #2076, opened 2021-09-13
CmdLet Copy-AlSourceFiles can't be used on BC 19x


RonKoppelaar:
When using Create-AlProjectFolderFromNavContainer from 19x preview build it fails with following error:

Copying Al Source Files from C:\ProgramData\BcContainerHelper\Extensions\Original-19.0.29956.0-NL-al\* to C:\ProgramData\bccontainerhelper\BaseApp
WARNING: Unexpected error while handling BusinessManagerEvaluationRC.PageCust.al
New-cdsaALProject : Exception calling "Substring" with "2" argument(s): "Length cannot be less than zero.
Parameter name: length"

root cause is the cmdlet: Copy-AlSourceFiles
Which can't handle object type "pagecustomization" which doesn't have an object ID

See code below
```
if ($alFileStructure) {
        $types = @('enum', 'page', 'table', 'codeunit', 'report', 'query', 'xmlport', 'profile', 'dotnet', 'enumextension', 'pageextension', 'tableextension', 'interface', 'entitlement', 'permissionset', 'permissionsetextension')
...
...
if ($type -eq "profile" -or $type -eq "interface" -or $type -eq "entitlement") {
````

RonKoppelaar:
@freddydk  just added a PR to fix this issue... Not tested though..

RonKoppelaar:
Tested my changes, which worked fine. Now I'm stubbling into another problem, it relates to deleted files.
See screenshot. They removed all the code but left the file pending there for historical reason?
For now I guess will do some workaround to remove those files.

![image](https://user-images.githubusercontent.com/33791875/133213719-cc1be1a0-a3a2-4ba1-b188-3b5111323b6b.png)

freddydk:
That is fixed in the fix I submitted

freddydk:
Shipped in 2.0.16

PR #2077, opened 2021-09-13
Add support for objecttype PageCustomization

RonKoppelaar:
#2076


freddydk:
Strange characteres in the file?
Also, there are a few other things that needs fixing for 19 - will fix myself.
Thanks for this though.

PR #2089, opened 2021-09-24
Further correction of 'if'

Silentapproach:
Sorry, there were two more. :) Did not see them at first.


freddydk:
Thanks, some copy/pasting going on there:-)

PR #2093, opened 2021-09-28
Improve Upgrade Handling while publishing an App

ChrisBlankDe:
Add Documentation for upgrade parameter.
If upgrade is not needed just install the app. this helps when the user just wants to have the app running but doesn't know if an upgrade is necessary or not. just always publish with the upgrade parameter and then the app is ready in any case.

### DIFF
--- a/AppHandling/Publish-NavContainerApp.ps1
+++ b/AppHandling/Publish-NavContainerApp.ps1
@@ -18,6 +18,8 @@
   Specify Add, Clean or Development based on how you want to synchronize the database schema. Default is Add
  .Parameter install
   Include this parameter if you want to install the app after publishing
+ .Parameter upgrade
+  Include this parameter if you want to upgrade the app after publishing. if no upgrade is necessary then its just installed instead.
  .Parameter tenant
   If you specify the install switch, then you can specify the tenant in which you want to install the app
  .Parameter packageType
@@ -319,7 +321,13 @@ try {
                             Sync-NavTenant -ServerInstance $ServerInstance -Tenant $tenant -Force
                             Sync-NavApp -ServerInstance $ServerInstance -Publisher $appPublisher -Name $appName -Version $appVersion -Tenant $tenant @syncArgs -force -WarningAction Ignore
                         }
-                
+
+                        $navAppInfoFromDb = Get-NAVAppInfo -ServerInstance $ServerInstance -Publisher $appPublisher -Name $appName -Version $appVersion -Tenant $tenant -TenantSpecificProperties
+                        if($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version -and $upgrade){
+                            $install = $true
+                            $upgrade = $false
+                        }
+                        
                         if ($install) {
         
                             $languageArgs = @{}
--- a/AppHandling/Publish-NavContainerApp.ps1
+++ b/AppHandling/Publish-NavContainerApp.ps1
@@ -322,10 +322,11 @@ try {
                             Sync-NavApp -ServerInstance $ServerInstance -Publisher $appPublisher -Name $appName -Version $appVersion -Tenant $tenant @syncArgs -force -WarningAction Ignore
                         }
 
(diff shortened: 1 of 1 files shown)


freddydk:
I understand the scenario, but if I select -upgrade I actually want it to fail if the prior app wasn't installed correctly or I accidently changed appid or other strange things.
We could do the following:
change this:
`if($navAppInfoFromDb.ExtensionDataVersion -eq $navAppInfoFromDb.Version -and $upgrade){
`to this
`if($navAppInfoFromDb.ExtensionDataVersion -eq $navAppInfoFromDb.Version -and $upgrade -and $install){
`
basically meaning that if you supply both parameters, then it will install or upgrade.
(and then remove the install=true)

Make sense?

ChrisBlankDe:
you are absolutely right. code adjusted.

freddydk:
Looking at it, I would like to request one more change in order to minimize impact on older versions etc.
this:
```
$navAppInfoFromDb = Get-NAVAppInfo -ServerInstance $ServerInstance -Publisher $appPublisher -Name $appName -Version $appVersion -Tenant $tenant -TenantSpecificProperties
if($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version -and $upgrade -and $install){
    $install = $true
    $upgrade = $false
}
```
should be 
```
if($upgrade -and $install){
    $navAppInfoFromDb = Get-NAVAppInfo -ServerInstance $ServerInstance -Publisher $appPublisher -Name $appName -Version $appVersion -Tenant $tenant -TenantSpecificProperties
    if($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version -and $upgrade -and $install){
        $install = $true
        $upgrade = $false
    }
}
```
Sorry for not including this feedback in the first round

freddydk:
Thanks

ChrisBlankDe:
hi freddy the optimizations were not so good after all. now the code doesn't do what it's supposed to.
if an old version was already installed and an upgrade is necessary `$install` is no longer set to `$false` and therefore runs into an error.
how do we proceed here? should i provide a pr with a fix?

freddydk:
If I read the code - install is set to $false if the app is already installed in the same version - install should not be set to false if an old version exists - or?
if you supply -ignoreIfAppExists - does it then work?

ChrisBlankDe:
you are right.
we have just tested a case which we had not considered: the old app version was already uninstalled.

why not just change it like this way: 

```powershell
    if($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version){
        $upgrade = $false
+   } else {
+       $install = $false
    }

```

jeesus450:
I have a new problem with this code:

```
if($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version){
        $upgrade = $false
+   } else {
+       $install = $false
    }
```

If -install and -upgrade are specified and there was no previous version of the app ( what means install), then after the sync $navAppInfoFromDb.ExtensionDataVersion = $null and does not equal $navAppInfoFromDb.Version what means $install = $false what is wrong.

I had to do 2 changes to this to make it work:

```
if(($navAppInfoFromDb.ExtensionDataVersion -eq  $navAppInfoFromDb.Version) -or `
($navAppInfoFromDb.ExtensionDataVersion.Length -eq 0))
{
     $upgrade = $false
} else {
     $install = $false
}
```

and I had to add $upgrade = $false under ignoreifinstalled
```
$appInfo = (Get-NAVAppInfo -ServerInstance $serverInstance -Name $navAppInfo.Name -Publisher $navAppInfo.Publisher -Version $navAppInfo.Version @addArg)
                        if ($appInfo) {
                            $publishIt = $false
                            Write-Host "$($navAppInfo.Name) is already published"
                            if ($appInfo.IsInstalled) {
                                $install = $false
                                Write-Host "$($navAppInfo.Name) is already installed"
                               ########## new ##########
                                $upgrade = $false
                                Write-Host "$($navAppInfo.Name) is already upgraded"
                               #######################
                            }
                        }
 
```

Sorry I did not have time to create a pull request, but it seems to work like this.

freddydk:
if you could copy/paste this into a new issue, then I will have a look when I return home from vacation.

PR #2155, opened 2021-10-25
Addded option for keyVaultUrls in Run_AlPipeline

ronaldnl:
This adds the option to add or overide keyVaultUrls to app.json when using Run-AlPipeline.
This is needed when keyVaultUrls need to be changed or added on building the app.

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -62,6 +62,8 @@
   ApplicationInsightsKey to be stamped into app.json for all apps
  .Parameter applicationInsightsConnectionString
   ApplicationInsightsConnectionString to be stamped into app.json for all apps
+ .Parameter keyVaultUrls
+  Array or comma separated list of keyVaultUrls to be stamped into app.json for all apps
  .Parameter testResultsFile
   Filename in which you want the test results to be written. Default is TestResults.xml, meaning that test results will be written to this filename in the base folder. This parameter is ignored if doNotRunTests is included.
  .Parameter bcptTestResultsFile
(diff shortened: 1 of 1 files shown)


freddydk:
QQ - why wouldn't you just add the KeyVaultUrl to the app.json in the repo?

ronaldnl:
We have for dev, but we use a different vault between dev and production version of our apps.
So right now I would have to stamp it in before using the run-alpipeline command in a seperate task. So I tought it would be nicer to change it at the point where other values of app.json are also getting updated.

And I did check if there was an override block for the app.json changes during build but I couldn't find it.

freddydk:
The issue here is that you would stamp the same KeyVaultUrl into all apps (library apps, main apps etc.)
Maybe a better solution is exactly what you were looking for - an override function you could specify where you can modify app.json Just-In-Time.
This could also solve the case, where people want to avoid Application Insights to be present during normal builds - and maybe other things.
Will think about this and get back, thanks

ronaldnl:
> The issue here is that you would stamp the same KeyVaultUrl into all apps (library apps, main apps etc.) Maybe a better solution is exactly what you were looking for - an override function you could specify where you can modify app.json Just-In-Time. This could also solve the case, where people want to avoid Application Insights to be present during normal builds - and maybe other things. Will think about this and get back, thanks

Thank you for the reply!
I can agree on the points you make. 

Would a similar setup like with Function UpdateLaunchJson work?
So a function before even creating the container which can be used on set parameters and/or an override script block. Since you could consider updating the app.json as a sort of initializing that should happen even before creating the container as in a way it is some sort of configuration.

If yes I could make a change for you and save you possibly some time. Which was my initial goal anyway of creating a contribution instead of just raising a question.

Also not sure if the other present app.json should be moved with them. In my head it would make at least sense to have all the changes to the app.json together. If yes I will move those as well.

freddydk:
Yes, thanks
And both functions should take a hashtable as parameter.
Note that when you read a .json you get a PSCustomObject - you can pipe that into ConvertTo-HashTable (Easier to work with).

PR #2169, opened 2021-11-03
initial support for traefik v2

tfenster:
Support for using Traefik v2

### DIFF
--- a/BcContainerHelper.psm1
+++ b/BcContainerHelper.psm1
@@ -67,6 +67,7 @@ function Get-ContainerHelperConfig {
                 "za" = "w1"
             }
             "TraefikUseDnsNameAsHostName" = $false
+            "forceTraefikV1" = $false
             "TreatWarningsAsErrors" = @('AL1026')
             "TelemetryConnectionString" = ""
             "UseExtendedTelemetry" = $false
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -776,7 +776,7 @@ try {
 
     if ($clickonce) {
         if ($useTraefik) {
-            Write-Host "WARNING: ClickOnce doesn't work with traefik v1 (which is the one used in this version of ContainerHelper)"
+            Write-Host "WARNING: ClickOnce doesn't work with traefik"
         }
         $parameters += "--env clickonce=Y"
     }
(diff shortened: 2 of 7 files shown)


tfenster:
@freddydk I finally got around to giving Traefik v2 support a try. This is completely untested, so I am pretty certain this doesn't work yet, but could you please give me a quick feedback whether you like the approach or I should do it differently? 

Forcing the usage of v1 is still possible through the config file (see https://github.com/microsoft/navcontainerhelper/pull/2169/files?diff=unified&w=1#diff-9d384d29ea5c25854f913091c0066e31f70a6b9e3c45834cb5261e08c876ccc8). I could also try for exiting environments to find out which Traefik version is used by checking the image tag of the running traefik container. Otherwise people with existing environments would need to make this config setting when updating bccontainerhelper.

There are also some whitespace changes in the PR but I don't have a clue where they are coming from or how I could fix them, sorry :(

freddydk:
The whitespace changes probably are line endings - no worries.
A few things

Right now, I am working on supporting Windows Server 2022 in the ARM templates.

The current traefik container doesn't work on server 2022 (built for 1809) and even if I run it in hyperv isolation - it fails.
Instead, when I rebuild the traefik image based on mcr.microsoft.com/windows/servercore:10.0.20348.288 - then it works.
the dockerfile is here: https://github.com/traefik/traefik-library-image/blob/master/windows/1809/tmplv1.Dockerfile and then replace ${VERSION} with v1.7.33 - then it works.

It also has the advantage that I reuse the same servercore with the subsequent BC containers (save 5Gb download)
I am thinking of building the image as part of the ARM template - what do you think about that idea?

This would however mean, that instead of having a -forcev1 in the bccontainerhelper config, we should have the imagename. And you would then have to look at the version label in the image to see whether is starts with v1 or v2 to determine the which settings to use.

So, in bccontainerhelper config: TraefikImage (set to the default traefik version) - and then have a parameter on Setup-traefik function called TraefikImage, which defaults to $bccontainerhelperconfig.traefikImage.
This has the advantage that people can specify the image on the call - or in the config.

Make sense?

In order for me to get the ARM templates to work - I will do the config changes in my dev branch today and test them out - you can grab the config change there to avoid conflict when merging later.

BTW - the code for determining traefik images doesn't seem to work:
docker images --filter "label=org.label-schema.name=Traefik" --format "{{.ID}}"
maybe it should be:
docker images --filter "label=org.opencontainers.image.vendor=Traefik Labs" --format "{{.ID}}"

freddydk:
https://aka.ms/getbc2 and https://aka.ms/getbcext2 (2 means grab from the dev branch of nav-arm-templates) now supports Windows Server 2022 and Windows 11 hosts.
For these - it builds the traefik container on the fly using the same server core image as BC and it seems to work just fine.
Will run a script now, which create 4 different VMs, with and without traefik for each supported OS == 32 VMs - and then test whether they all work.

freddydk:
Oh, and BTW - the new arm templates now use Docker Engine instead of msftprovider for server and docker desktop for desktop.

tfenster:
I won't be able to do this anymore as we are moving away from Traefik in COSMO Alpaca, so I don't have (professional) contact with Traefik anymore. If anyone wants to do something similar, feel free to reach out for feedback and review if you want

PR #2194, opened 2021-11-16
Update Remove-NavContainerSession.ps1

vody:
Currently, sessions will not be evicted correctly as sessions collection using 'containerName' as a reference in [Get-NavContainerSession.ps1](https://github.com/microsoft/navcontainerhelper/blob/master/ContainerHandling/Get-NavContainerSession.ps1). This small fix will use 'containerName' as a reference to remove sessions from sessions collection.


freddydk:
You are right - and this bug has existed since July 21st 2020...:-(
Thanks

PR #2198, opened 2021-11-17
Added new parameter to New-NavContainerNavUser.ps1

JeppeBylov:


freddydk:
Instead of adding fullname directly to the functions, you should add it to $parameters conditionally (like I do with the authenticationEmail).
This ensures that the outer function behaves exactly like before if the parameter is not specified.

freddydk:
Thanks Jeppe

PR #2202, opened 2021-11-23
add param containerName

CriticalMax:
Hey Freddy,

seems like this parameter is neccessary. Containers were able to be created but were missing the test toolkit. It's a minor fix but a new release would be appreciated!


freddydk:
Thanks - just saw issue #2206  and fixed this with the same fix.

freddydk:
fixed in 2.0.22

PR #2232, opened 2021-12-13
Remove Windows Users from a database

jonaswre:
When exporting a database to a bacpac its not allowed for windows users to be present in the database. 

This is no problem when the database is newly created within the container. 
If a bak file is used when creating the container. Its possible that there are users present that cant be exported to bacpac.

One option to solve the issue is to remove the users ether after import or before the export.

Since some users might want to keep exporting baks that still have the users. I thinks its better to remove them in the temp database before exporting to bacpac.


freddydk:
Are Windows Users only in the application database - wouldn't you need to remove it from tenant db as well?
Wouldn't you call your new function right after the call to Remove-NetworkServiceUser (in Remove-NavTenantDatabaseUserData and Remove-NavDatabaseSystemTableData)?

beside that, please add a line to release notes text file in the root under 3.0.1, thanks

jonaswre:
You are right, the windows users should be cleared from any database before export. I've added the change in the latest commit

jonaswre:
@freddydk is this now acceptable?

freddydk:
Yes, thanks

Issue #2237, opened 2021-12-17
New-BcContainer reports that docker is not running besides docker being installed and running.


Sven-Niehus:
**Describe the issue**
New-BcContainer reports that docker is not running besides docker being installed and running. 

The issue seems to be reported, because New-BcContainer checks for a hardcoded process called `docker` to determine if docker is running. [Source.](https://github.com/microsoft/navcontainerhelper/blob/fd2ae4841b0fd8b6ad2abd465d4df7ef9f2e79db/ContainerHandling/New-NavContainer.ps1#L475-L478)
A fix would be looking for a running `dockerd.exe` process or run something like `docker ps` to check if it will return an error.

I have already created a PR with the first fix: #2238 

**Scripts used to create container and cause the issue**
```
$containerName = 'bcserver'
$credential = Get-Credential -Message 'Using UserPassword authentication. Please enter credentials for the container.'
$auth = 'UserPassword'
$artifactUrl = Get-BcArtifactUrl -type 'OnPrem' -country 'de' -select 'Latest'
New-BcContainer `
    -accept_eula `
    -containerName $containerName `
    -credential $credential `
    -auth $auth `
    -artifactUrl $artifactUrl `
    -updateHosts
```

**Full output of scripts**
```
BcContainerHelper version 3.0.0
BcContainerHelper emits usage statistics telemetry to Microsoft
BcContainerHelper is version 3.0.0
BcContainerHelper is running as administrator
Host is Microsoft Windows Server 2022 Standard Evaluation - ltsc2022
New-BcContainer Telemetry Correlation Id: b11073d7-77bc-4e7f-b60f-52ba7b70a568
Docker Service not found. Docker is not started, not installed or not running Windows Containers.
At C:\Program Files\WindowsPowerShell\Modules\BcContainerHelper\3.0.0\ContainerHandling\New-NavContainer.ps1:477 char:9
+         throw "Docker Service not found. Docker is not started, not i ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Docker Service ...ows Containers.:String) [], RuntimeException
    + FullyQualifiedErrorId : Docker Service not found. Docker is not started, not installed or not running Windows Containers.

```

**Screenshots**

**Additional context**
```
PS > docker images
REPOSITORY                          TAG              IMAGE ID       CREATED        SIZE
[...]

PR #2238, opened 2021-12-17
Look for a running "dockerd" process and not a running "docker" service to determine if docker is running

Sven-Niehus:
See issue #2237

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -472,7 +472,7 @@ try {
 
     Write-Host "Host is $($os.Caption) - $hostOs"
 
-    $dockerService = (Get-Service docker -ErrorAction Ignore)
+    $dockerService = (Get-Process "dockerd" -ErrorAction Ignore)
     if (!($dockerService)) {
         throw "Docker Service not found. Docker is not started, not installed or not running Windows Containers."
     }


freddydk:
changing this would cause all other machines to fail:
image
I would rather like to know: How did you install docker? what SKU are you running?

Sven-Niehus:
> changing this would cause all other machines to fail:

In the proposed change I'm looking for a Process called dockerd (```Get-Process "dockerd" -ErrorAction Ignore```) not for a Service with the name ```dockerd```. The process dockerd should be present on all machines, because the docker daemon needs to be running for everything with docker.

> I would rather like to know: How did you install docker? what SKU are you running?

As I'm on Windows Server I have installed docker via the binaries from [here](https://download.docker.com/win/static/stable/x86_64/) as you mentioned in [this](https://freddysblog.com/2021/10/30/docker-and-business-central/) blog post. After downloading the binaries I have installed the docker service via [nssm](https://github.com/kirillkovalenko/nssm) and named the service ```dockerd```

freddydk:
good point, thanks

freddydk:
This fix broke containerhelper tests.
The following lines checked the status on the dockerservice - and the process object doesn't have a status.
Not sure how the PR could ever work?

I have removed the following lines and resubmitted.

Issue #2236, opened 2021-12-16
Need a way to know the creation date of BcContainerTenants


vvkjndl:
**Describe the issue**
We have a requirement to remove tenants that were created before a specific date in a container running 100+ tenants.

Playing around with functions within BcContainerHelper, I wasn't able to find any way to determine the creation date of the tenant.

**Scripts used to create container and cause the issue**
```
Get-BcContainerTenants -containerName test -tenant default
```

**Full output of scripts**
```
PSComputerName                         : c8984ac2fa7fc4e06b33963191d4651714eebcbf7d05232b8073cd10adc1783a
RunspaceId                             : fcb9d548-0e6c-48c8-b3c8-b462d5270911
ServerInstance                         : MicrosoftDynamicsNavServer$BC
DatabaseName                           : default
DatabaseServer                         : localhost\SQLEXPRESS
DatabaseUserName                       :
State                                  : Operational
DetailedState                          :
DeletionState                          : Not deleted
IsInExclusiveAccessMode                : False
TenantDataVersion                      : 19.3.33850.0
Compression                            : Page
Id                                     : default
AlternateId                            : {vj-default}
AllowAppDatabaseWrite                  : False
NasServicesEnabled                     : False
RunNasWithAdminRights                  : False
EncryptionProvider                     : LocalKeyFile
AzureKeyVaultSettings                  :
DefaultCompany                         :
DefaultTimeZone                        : UTC
ExchangeAuthenticationMetadataLocation :
AadTenantId                            :
ApplicationInsightsConnectionString    :
DisplayName                            :
EnvironmentName                        :
EnvironmentType                        : Sandbox
EnvironmentSettings                    :
```

**Screenshots**
N/A

**Additional context**
- does it happen all the time?
N/A

- did it use to work?
N/A

freddydk:
I don't think you can see that with any BC cmdlets.
Maybe you can see when the tenant database was created - https://sqlandme.com/2013/02/19/sql-server-finding-out-database-creation-time/

PR #2240, opened 2021-12-21
Add size and datecreated information to Get-BcContainerTenants cmdlet

vvkjndl:
Add size and datecreated information to Get-BcContainerTenants cmdlet


vvkjndl:
#2236

freddydk:
I think this functionality is very specialized.
Instead of adding this to containerhelper, you should just have a function in your own library that does this, Invoke-ScriptInBcContainer is helpful (as you know)

In the future, BcContainerHelper will be rewritten to PS7 and we will try to remove some of the very specialized behaviors to make support easier. Therefore I think it is wrong to add more now.

vvkjndl:
Noted, this can be canceled. I have already built a function for my own usage.

PR #2289, opened 2022-02-02
App is only ignored from resolution if it has a dependecies property in the app.json 

jonaswre:
If I understand correctly. Having a "dependencies" property in the app.json is not mandatory. 

If the property is missing the following block will not be executed.

```
if ($appJson.psobject.Properties.name -contains "dependencies") {
  if ($appJson.psobject.Properties.name -contains "id" -and $lvl -eq 0) {
      $ignoredDependencies += $appJson.Id
  }
  ...
}
```

This will result in the app being downloaded even if it shouldn't.

Which might result in an error in the CI/CD Pipeline.


freddydk:
LGTM

PR #2295, opened 2022-02-04
App Resolutions not working correctly with lower runtimes.

jonaswre:
In runtime 4.0 and lower the app id is present with the the property appId and not id.

https://github.com/microsoft/navcontainerhelper/blob/89c424816ab8c4ae6cfd5f903d50cb24da00cb4e/AppHandling/Extract-AppFileToFolder.ps1#L165-L176


freddydk:
Thanks

PR #2316, opened 2022-02-16
Added blobPath to Publish-BuildOutputToStorage 

### LINKED ISSUE #2308: Possibility to define a base path when publishing to Azure Storage
**Describe the issue**
When publishing a package to Azure Storage, it would be helpful to have the ability to define a base folder/path for the app inside the container. 

At this moment there is only the possibility to define the Container name. If there is the need to organize the storage in a more structured way, i.e. Project (Container) > Folder > version/latest (Folder), this is not possible. This is particulary useful in a case where in the same Project there is multiple repositories (each representing an app).

I would suggest to add a new parameter (optional) to Publish-BuildOutputToStorage, where it would be possible to define a basePath and change the calls for Set-AzureStorageBlobContent to include it before ""$appVersion/$_.zip".ToLowerInvariant()" and ""latest/$_.zip".ToLowerInvariant()".

Thank you!

jonaswre:
This resolves #2308


freddydk:
There is a difference in functionality here.

My pipelines use this with setLatest - in order to store the version BOTH by version number AND as Latest.
When the next version is released, then the same happens, but only the Latest is overridden.
With the new fix - the blob is only stored as Latest and the old version will no longer be available.

PackageHandling/Publish-BuildOutputToStorage.ps1
```diff
@@ -58,22 +61,33 @@ try {
             $tempFile = Join-Path (Get-TempDir) "$([Guid]::newguid().ToString()).zip"
             try {
                 Compress-Archive -path (Get-Item (Join-Path $path $_)).FullName -DestinationPath $tempFile
-                Set-AzureStorageBlobContent -File $tempFile -Context $storageContext -Container $projectName -Blob "$appVersion/$_.zip".ToLowerInvariant() -Force | Out-Null
+
+                $version = $appVersion
```
freddydk, line 65:
```
$versions = @($appVersion)
if ($setLatest) { $versions += @("Latest") }
$versions | % {
   set blob
}
`

jonaswre: Yes you are right. setLatest should still upload in a version directory

freddydk:
LGTM

Issue #2351, opened 2022-03-17
Plans for Support of MsGraph instead of AzureAD


pri-kise:
**Describe the issue**
I'm trying to setup new containers with Aad Authentitaciton.
Therefore we are want to make use of a azure app registration that has the required permission to create new applications for the business central docker instances.

I try to follow the nice tutorial of Tobias Fenster (https://tobiasfenster.io/using-aad-authentication-for-development-against-a-business-central-container) with one difficulty. I cannot add any of the `Azure Active Directory Graph` permissions to new app registrations.

I know that there is the new Powershell Module to make use of the MsGraphApi instead (https://github.com/microsoftgraph/msgraph-sdk-powershell).

I now have the problem that I cannot get an AadAccesstoken with the required permissions for the BCContainerHelper Commands since those commands for creating a app and users are still using the old AzureAd Authentication.

Is there anything planned to support both variants in the future?

I can try to clarify my requests, if it isn't clear what I ment.

**Scripts used to create container and cause the issue**
```
```

**Full output of scripts**
```
```

**Screenshots**

**Additional context**

freddydk:
You should be able to use New-BcAuthContext to get a BcAuthContext, which you can give to Create-AadAppsForBC
New-BcAuthContext can be called with a refreshtoken or using -includedevicelogin and other mechanisms to get access.
I don't have any plans of re-writing that function just yet - eventually it will probably have to be.,

freddydk:
https://freddysblog.com/2021/01/25/bcauthcontext/

pri-kise:
Hi thanks for you fast Reply.
I tried to get it to work with BcAuthContext, but I'm having some problems.

```al
Import-Module BcContainerHelper
$azureApplicationId ="00000000-0000-0000-0000-000000000000"
$azureTenantId = "00000000-0000-0000-0000-000000000000"
$azurePassword = ConvertTo-SecureString "xxx" -AsPlainText -Force

$bcAuthContext = New-BcAuthContext -tenantID $azureTenantId -clientID $azureApplicationId -clientSecret $azurePassword -scopes "https://graph.windows.net/.default"

$containerName = "test23"

$publicDns = "https://xxx.de"
$publicWebBaseUrl = "$publicDns/$containerName"

$appIdUri = $publicWebBaseUrl

$bcContainers = Get-BcContainers -includeLabels | Where-Object name -EQ $containerName
if ($bcContainers.Count -ne 0) 
{
[...]

PR #2366, opened 2022-03-23
New Cmdlt New-AadAppsForBc using Microsoft.Graph

pri-kise:
I tried to add new cmdlet that supports the new Microsoft.Graph Powershell Module to create the Aad  Apps. (See issue https://github.com/microsoft/navcontainerhelper/issues/2351)

it can be tested with the following commands:
```ps
$bcAuthContext = New-BcAuthContext -tenantID $azureTenantId -clientID $azureApplicationId -clientSecret $clientSecret -scopes "https://graph.microsoft.com/.default"
  $AdProperties = New-AadAppsForBc -appIdUri https://mycontainer.mydomain/bc/ -bcAuthContext $bcAuthContext 
```
If there is anything that should be changed, I'm happy to do so. I'm no expert with powershell, therefore there can be some room for improvements.

### DIFF
--- a/AzureAD/New-AadAppsForBc.ps1
+++ b/AzureAD/New-AadAppsForBc.ps1
@@ -197,7 +197,7 @@ try {
 
     if ($preAuthorizePowerShell) {
         $PreAuthorizedApplications = $ssoAdApp.Api.PreAuthorizedApplications
-        $PreAuthorizedApplications += @{ "AppId" = "1950a258-227b-4e31-a9cf-717495945fc2"; "PermissionIds" = @($oauth2permissionid) }
+        $PreAuthorizedApplications += @{ "AppId" = "1950a258-227b-4e31-a9cf-717495945fc2"; "DelegatedPermissionIds" = @($oauth2permissionid) }
         Update-MgApplication -ApplicationId $ssoAdApp.Id -Api @{PreAuthorizedApplications = $PreAuthorizedApplications}
     }
 
--- a/AzureAD/New-AadAppsForBc.ps1
+++ b/AzureAD/New-AadAppsForBc.ps1
@@ -1,0 +1,366 @@
+﻿<# 
+ .Synopsis
+  Create Apps in Azure Active Directory to allow Single Signon when using AAD
+ .Description
+  This function will create an app in AAD, to allow Web and Windows Client to use AAD for authentication
+  Optionally the function can also create apps for the Excel AddIn and/or PowerBI integration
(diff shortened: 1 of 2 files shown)


freddydk:
Thanks. I will go through this after Directions NA

pri-kise:
@freddydk I'm really looking forward for your feedback on this new cmdlet that it'll be available before the [EOL of the  Azure AD API](
https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363) and that someone can find bugs and improve this cmdlt.

freddydk:
Thanks - will do - I will need to be able to set aside a block of time for this.
Thanks a lot for your submission and rest assured that this will be fixed in good time before EOL of AAD API.

freddydk:
Thanks, will try to get this done next week.

pri-kise:
@freddydk just a short reminder for this PR. 
Maybe you'll find some time in the new year to take a look at this specific scenario to enable support the MS Graph API.

freddydk:
I definitely will

freddydk:
This PR fixes a few issues with the functions, which I ran into - and adds them to the exported functions.
https://github.com/microsoft/navcontainerhelper/pull/2996

It also adds a deprecation warning to the Create-AadAppsForBc.

Let me know if you find any issues with my changes.

pri-kise:
@freddydk I've checked our source and we added the following to grant permission access to the new created application.

```ps
    $client = Get-MgServicePrincipal -Filter "appId eq '$appId'"
    $resource = Get-MgServicePrincipal -Filter "servicePrincipalNames / any(n: n eq 'https://graph.microsoft.com')"
    New-MgOAuth2PermissionGrant -ClientId $client.Id -ConsentType "AllPrincipals" -ResourceId $resource.Id
```
Maybe this can be simplified  or improved but I only want to the leave this here for reference.

freddydk:
Thanks, so this will avoid that when I connect the first time, I need to consent?

pri-kise:
Exactly. The first approval was very annoying for our developers and we wanted to avoid this step.
I don't know if the app registration (that we use for authentication) needs specific rights to perform this request.
See: https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.identity.signins/new-mgoauth2permissiongrant?view=graph-powershell-1.0
Maybe the client id and the resource Id could retrieved more easily or save during the creation process and reused afterwards.

PR #2405, opened 2022-04-12
allow whitespaces around IP and Hostnames in hosts file.

thellMa:
Had issues with building a new container with use of external SQL.  Found out that I had some whitespaces in my hosts file.
This resolves the issue.

handles following example lines in hosts file.  (without " of course)
"    192.168.1.190  host.docker.internal  "
"192.168.1.190          host.docker.internal"


freddydk:
Could you add a line in the releaseNotes.txt file under 3.0.5 - then I will merge it.

freddydk:
Thanks

PR #2426, opened 2022-04-28
Add appFilePath param to Get-BcContainerAppInfo

rvanbekkum:
This PR adds an `-appFilePath` parameter to the `Get-BcContainerAppInfo` function.
This parameter is passed as the `-Path` parameter for the `Get-NAVAppInfo` CmdLet.
For this parameter the `-ServerInstance` parameter should not be passed (this would result in an error).

We were currently missing this parameter set as an option in the BcContainerHelper module.
Please let me know if you have any comments on the implementation.

### DIFF
--- a/AppHandling/Get-NavContainerAppInfo.ps1
+++ b/AppHandling/Get-NavContainerAppInfo.ps1
@@ -2,7 +2,7 @@
  .Synopsis
   Get App Info from NAV/BC Container
  .Description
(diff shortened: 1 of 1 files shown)


rvanbekkum:
@freddydk Was this accidentally closed by deleting the `dev` branch?
Should this PR target `master` instead?

freddydk:
My bad - sorry

PR #2431, opened 2022-04-29
Add appFilePath param to Get-BcContainerAppInfo

rvanbekkum:
This PR adds an `-appFilePath` parameter to the `Get-BcContainerAppInfo` function.
This parameter is passed as the `-Path` parameter for the `Get-NAVAppInfo` CmdLet.
For this parameter the `-ServerInstance` parameter should not be passed (this would result in an error).

We were currently missing this parameter set as an option in the BcContainerHelper module.
Please let me know if you have any comments on the implementation.

(Opening a new PR as https://github.com/microsoft/navcontainerhelper/pull/2426 was closed by @freddydk)

### DIFF
--- a/AppHandling/Get-NavContainerAppInfo.ps1
+++ b/AppHandling/Get-NavContainerAppInfo.ps1
@@ -51,7 +51,8 @@ try {
 
     $args = @{}
     if ($appFilePath) {
-        $args += @{ "Path" = $appFilePath }
+        $containerAppFilePath = Get-BcContainerPath -containerName $containerName -path $appFilePath -throw
+        $args += @{ "Path" = $containerAppFilePath }
     }
     elseif ($symbolsOnly) {
         $args += @{ "SymbolsOnly" = $true }
--- a/AppHandling/Get-NavContainerAppInfo.ps1
+++ b/AppHandling/Get-NavContainerAppInfo.ps1
@@ -2,7 +2,7 @@
  .Synopsis
   Get App Info from NAV/BC Container
  .Description
-  Creates a session to the NAV/BC Container and runs the CmdLet Get-NavAppInfo in the container
+  Creates a session to the NAV/BC Container and runs the CmdLet Get-NAVAppInfo in the container
  .Parameter containerName
   Name of the container in which you want to enumerate apps
  .Parameter tenant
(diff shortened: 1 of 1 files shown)


AppHandling/Get-NavContainerAppInfo.ps1
```diff
@@ -15,30 +15,45 @@
...
         [switch] $installedOnly
     )
 
 $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
 try {
 
     $args = @{}
-    if ($symbolsOnly) {
+    if ($appFilePath) {
+        $args += @{ "Path" = $appFilePath }
```
freddydk, line 54: I think you need to use 
"Path" = Get-BcContainerPath $appFilePath
in order for this to work (unless the path is the same inside and outside the container)

freddydk (changes requested):

AppHandling/Get-NavContainerAppInfo.ps1
rvanbekkum: I think you might be right, I think I may have tested with a scenario where it works regardless.
I have made adjustments accordingly.

freddydk:
Thanks Rob will be in 3.0.8

TomHedegaardTrimit:
Error "A positional parameter cannot be found that accepts argument XXXXX"..
When running data test in our pipeline, we suddenly got an error saying that a positional parameter was missing, but it did not say anything about where and why. We have not changed anything in the pipeline script since in ran successfully last time.

After trying several things in the script, I finally found the line that caused the problem and solved it.

In the test pipeline script, we use this command (which gave the error):
$testApps = @(Get-NavContainerAppInfo $containerName | Where-Object { $_.Name -like "Tests-*" })

I changed into this line:
$testApps = @(Get-NavContainerAppInfo -containerName $containerName | Where-Object { $_.Name -like "Tests-*" })

So, you might consider to change the behavior in containerhelper so that it is not required to use the -containerName, but allowe only the $containerName ?

image

freddydk:
@TomHedegaardTrimit - sorry for not being explicit.
If you can create a new issue with this, then I can reference that with in the release notes and other people can find it.
It is hard to find a comment on a PR.

PR #2465, opened 2022-05-16
fix issue with reportsuppresseddiagnostics

freddydk:


marknitek:
@freddydk imo this fix (or more precisely the whole check to only apply ReportSuppressedDiagnostics when alc supports it) should be applied to Run-AlCops.ps1 as well. 

We invoke Run-AlCops on the 20.0 Version  currently and get an error when doing so because the parameter is applied by default

freddydk:
Run-AlCops calls Compile-AppInBcContainer - this fix will make sure that is fixed as well.

marknitek:
Oops i missed that point, thanks for the fix!

PR #2488, opened 2022-05-20
Set-BcContainerServerConfiguration and Restart-BcContainerServiceTier added

DanielGoehler:
I added the `Set-BcContainerServerConfiguration` and `Restart-BcContainerServiceTier` cmdlets to make it easier to change the Service Tier configuration.

### DIFF
--- a/AppHandling/Get-NavContainerAppInfo.ps1
+++ b/AppHandling/Get-NavContainerAppInfo.ps1
@@ -28,6 +28,7 @@
 #>
 function Get-BcContainerAppInfo {
     Param (
+        [Parameter(Position=0)]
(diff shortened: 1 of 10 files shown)


freddydk:
Quick you are:-)

Please add "| Out-Null" after both Invoke-script commands (after the argumentlist) to avoid any un-intended output.
And move the releasenotes to 3.0.9 (which hasn't been released yet, so this will be included in that)

freddydk:
on 5. - when running with usepssession = $false - all output in the script (also output to host) will be added to the return value.
Hence | Out-Null is the safe bet.

DanielGoehler:
Thanks. Done.

DanielGoehler:
Someting went wrong let me try again.

PR #2513, opened 2022-05-31
replace internet explorer icons with default browser icons

louagej:


freddydk:
Thanks

if you put this code:

```
$ProgramId = (Get-ItemProperty HKCU:\Software\Microsoft\windows\Shell\Associations\UrlAssociations\http\UserChoice).Progid
$key = "HKLM:\SOFTWARE\Classes\$ProgramId\shell\open\command"
$exePath = (Get-ItemProperty -Path $key)."(default)" -replace " *--.*", ""
$exePath = $exePath.Replace("`"", "")
```
in a try/catch and in the catch, you set the exePath to the old iexplore default - then I will take this.

PR #2525, opened 2022-06-09
Fix forceHttpWithTraefik

sirhc101:
Currently `forceHttpWithTraefik` does not work correctly, because of the following reasons:

1. Public URLs in BC Configuration still pointing at `https://`
2. Traefik is still redirecting `http://` to `https://`
3. Traefik entry point (`traefik.frontend.entryPoints`) still using `https://`

This causes that the Container is up and running, but web client will not be loaded due to cross-site scripting limitations and still complaining about missing certificates, etc.

I fixed this by removing the redirection from traefik toml, using the `$traefikProtocol` (http or https) variable as `traefik.frontend.entryPoints` label and specify protocol for public URLs based and the `forceHttpWithTraefik` switch.

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -1662,7 +1662,11 @@ if ($multitenant) {
         $dlPart = "/${containerName}dl"
         $webclientPart = "/$containerName"
 
-        $baseUrl = "https://$publicDnsName"
+        if ($forceHttpWithTraefik) {
+            $baseUrl = "http://$publicDnsName"
+        } else {
+            $baseUrl = "https://$publicDnsName"
+        }
         $restUrl = $baseUrl + $restPart
         $soapUrl = $baseUrl + $soapPart
         $webclientUrl = $baseUrl + $webclientPart
@@ -1717,7 +1721,7 @@ if ($multitenant) {
                                    "-l `"traefik.dl.port=8080`"",
                                    "-l `"traefik.dl.protocol=http`"",
                                    "-l `"traefik.enable=true`"",
-                                   "-l `"traefik.frontend.entryPoints=https`""
+                                   "-l `"traefik.frontend.entryPoints=$traefikProtocol`""
         )
 
         ("
--- a/ContainerHandling/traefik/template_traefik.toml
+++ b/ContainerHandling/traefik/template_traefik.toml
@@ -13,8 +13,6 @@ endpoint = "npipe:////./pipe/docker_engine"
 [entryPoints]
   [entryPoints.http]
   address = ":80"
-    [entryPoints.http.redirect]
-    entryPoint = "https"
   [entryPoints.https]
   address = ":443"
   [entryPoints.https.tls]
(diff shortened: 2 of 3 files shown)


tfenster:
That is actually intended behavior. "forceHttp" should force the BC container to use http, althought that can cause issues with mobile client auth. But the traefik frontend should always be https. At least that was the intention.

sirhc101:
> That is actually intended behavior. "forceHttp" should force the BC container to use http, althought that can cause issues with mobile client auth. But the traefik frontend should always be https. At least that was the intention.

Did I get you right? 
BC Web Client, OData, SOAP, etc. should be available by HTTP (Port 80), but the traefik Dashboard should be served by HTTPS (Port 443)?

The current behavior, using `forceHttpWithTraefik`, is that you will be forwarded from HTTP to HTTPS. This causes that the Web Client simply not loads, because it tries to consume resources from a mixed HTTP/HTTPS source (cross-site scripting). I did not tried SOAP and ODATA in the original setup, but additionally VSCode is not working due to untrusted connection.

I can take a look at the traefik dashboard thing, but I would assume that this setup is not that much common, so I'm not sure, it worthy to invest into this? I just got into this (for the first time ever) because a customer wants to host all developer containers on one server using traefik, only available internally and does not own a certificate we could use. So Lets Encrypt does not work and for only internal access HTTP was enough imho.

tfenster:
No, the traefik "frontend" is what takes all requests coming to traefik (see https://doc.traefik.io/traefik/v1.7/basics/#frontends). And that in the current design is always intended to run through https. `forceHttpWithTraefik` should only influence the backend (see https://doc.traefik.io/traefik/v1.7/basics/#backends), the BC container in our case: Without that setting the container also runs with https with a self-signed cert, but if you for whatever reason require http in the container, you can use `forceHttpWithTraefik`. So assuming you are a "client", calling any of the services in the BC container, this is what it looks like without `forceHttpWithTraefik`:

client --> https --> traefik --> https --> container

With the setting, this is what happens:

client --> https --> traefik --> http --> container

Or at least that was the intention :) And I am pretty sure that I also validated it when creating it. The only exception to that rule is the download site, which is always http in the container, if I remember correctly.

As you want to make a call from the client with http, that would be a new feature and I wouldn't change the existing one as it breaks compatibility for those who are using it today. To be honest, I think it would be easier and cheaper to go the https route, with an official or self-signed certificate as you won't be able to use the mobile client and also web browsers are already showing insecurity warnings if you do http.

And on a small side note: I have support for traefik 2 on my to-do list for a veeery long time, but I hope to get that done over my summer vacation at the latest.

freddydk:
I never looked at this, but I think that the current traefik implementation is used pretty heavily, also with https, so why are we not seeing everybody complaining about this?

sirhc101:
Well, it's a pretty specific scenario, to run BC with http only and traefik. I needed it just once, for a customer.
Not sure if it is relevant for the most users

freddydk:
Got it - I think in this case, people can do the modifications manually and we will keep this as is.

PR #2567, opened 2022-06-28
Fixed Dependency bug in Get-BcContainerAppInfo 

jonaswre:
There is an Issue with the Get-BcContainerAppInfo. 

If you take a look at the Dependencies Property it seems like the dependency is converted toString(). This has the drawback, that you cannot access the id property. 

```json
...
  "Dependencies": [
    "Application, Microsoft, 20.0.0.0, False"
  ],
...
```

With this fix you will be able to pass a parameter `-full` to the command to convert the full object to json.

Adding the full parameter ensures backwards compatibility.

```json
...
  "Dependencies": [
    {
      "AppId": "c1335042-3002-4257-bf8a-75c898ccb1b8",
      "Name": "Application",
      "Publisher": "Microsoft",
      "MinVersion": {
        "Major": 20,
        "Minor": 0,
        "Build": 0,
        "Revision": 0,
        "MajorRevision": 0,
        "MinorRevision": 0
      },
      "CompatibilityId": {
        "Major": 1,
        "Minor": 0,
        "Build": 0,
        "Revision": 0,
        "MajorRevision": 0,
        "MinorRevision": 0
      },
      "IsPropagated": false
    }
  ],
 ...
```


freddydk:
I like this - and I think we should also create a setting in bccontainerhelper.psm1, which is default true:

GetBcContainerAppInfoReturnsFullResultSets: true

and then set the default value of Full to :$bccontainerhelperconfig.GetBcContainerAppInfoReturnsFullResultSets

This would break compatibility, but if anybody uses the string property, they can set this setting.

Makes sense?

freddydk:
Did the release notes disappear?

jonaswre:
@freddydk like this? 

But this change might be a bit dangerous. This doesn't not only change the Dependencies property but all the other as well. 

meaning app id needs to be accessed via $_.AppId.Value instead of $_.AppId

It would certainly break the code for almost everybody, that use is command. 
But it would be fine for me. I just wanted to document the danger fully. 
To be honest I was a bit suprised that it wasn't already working this way.  Because you are doing the exact same in the Sort-AppsByDependency.  But I guess thats the reason why you are not using the command.

freddydk:
Ahh - yeah that would be a problem...
I will investigate and get back

jonaswre:
Maybe just set it to $false and let people who need it opt in. 

I would be glad if this could be merged sooner than later, so that i dont have to depoly my own version again.

jonaswre:
I wonder why powershell can't handle json arrays... 

```
$a = @{ "a" = "1" }
$b = @( @{ "b" = "1" })

$a | ConvertTo-Json
ConvertTo-Json -InputObject $a

$b | ConvertTo-Json
ConvertTo-Json -InputObject $b

function FullConvertTo-Json {
        param
        (
            [Parameter(ValueFromPipeline=$true)]
            $inputObject
        )
        ConvertTo-Json -Depth 99 -InputObject $inputObject
       
}

$b | FullConvertTo-Json
```
i've just tried the following. Even that doesn't work.



The json with your script is not completely right either.

```json
 "Dependencies":  {
                             "value":  [
                                           "@{AppId=c1335042-3002-4257-bf8a-75c898ccb1b8; Name=Application; Publisher=Microsoft; MinVersion=; CompatibilityId=; IsPropagated=False}"
                                       ],
                             "Count":  1
                         },
```

Im not sure whats happening here. A powershell array represented as a object?

jonaswre:
Okay, it now works the same. 

Maybe im wrong, but Write-Host doesn't work with the dependency any more. Its just empty not even any type. But thats alright, i was just seeing wrong prints.

freddydk:
Let's try this out - thanks for pointing out this issue

jonaswre:
Thanks for your input

freddydk:
Not sure whether you saw this, but I moved the CI/CD pipelines to GitHub over the weekend.
This PR was commited here: https://github.com/microsoft/navcontainerhelper/actions/runs/2577170127
And published to preview.

kine:
@freddydk  What is the result? What I need to do to have old behavior? Since last update (3.0.12) pipelines are failing because Get-BcContainerAppInfo is returning different results (pipeline is trying to update app even when it is not installed).  From this conversation I am not sure, if the behavior is backward compatible or I need to set some switch. Thanks.

Issue #2581, opened 2022-07-07
Event log messages show the wrong description


mtaanquist:
**Describe the issue**
When attempting to inspect the Event Logs from a container, they all have the wrong description in the "General" tab of a log entry, while the XML shows the correct details.

**Scripts used to create container and cause the issue**
```
New-BcContainer bctest200 `
    -accept_eula `
    -alwaysPull `
    -artifactUrl (Get-BCArtifactUrl -type OnPrem -country DK -version 20) `
    -assignPremiumPlan `
    -auth NavUserPassword `
    -doNotExportObjectsToText `
    -enableTaskScheduler `
    -isolation process `
    -licenseFile "BC 20 DEV.flf" `
    -updateHosts 
```

**Full output of scripts**
```
BcContainerHelper version 3.0.11
BcContainerHelper emits usage statistics telemetry to Microsoft
BcContainerHelper is version 3.0.11
BcContainerHelper is running as administrator
Hyper-V is Enabled
UsePsSession is True
Host is Microsoft Windows 10 Enterprise - 21H2
Docker Client Version is 20.10.16
Docker Server Version is 20.10.16
Removing Session bctest200
Removing container bctest200
Removing entries from hosts
Removing bctest200 from container hosts file
Removing bctest200-* from container hosts file
Removing Desktop shortcuts
Removing C:\ProgramData\BcContainerHelper\Extensions\bctest200
Fetching all docker images
Fetching all docker volumes
INFO: Windows 10 21H1/21H2 images are not yet available, using 2004 as these are found to work better than 20H2 on 21H1/21H2
Pulling image mcr.microsoft.com/businesscentral:10.0.19041.1415
10.0.19041.1415: Pulling from businesscentral
Digest: sha256:aa9f9d0779654ab9dba736a43141591a70fa419c4a9a57fd8a517086841280ff
Status: Image is up to date for mcr.microsoft.com/businesscentral:10.0.19041.1415
mcr.microsoft.com/businesscentral:10.0.19041.1415
Using image mcr.microsoft.com/businesscentral:10.0.19041.1415
Creating Container bctest200
Style: onprem
Multitenant: No
Version: 20.2.41144.41423
Platform: 20.0.41086.41354
Generic Tag: 1.0.2.10
Container OS Version: 10.0.19041.1415 (2004)
Host OS Version: 10.0.19044.1766 (21H2)
Using process isolation
Using locale da-DK
[...]

PR #2583, opened 2022-07-08
Archive evtx file to preserve display information

mtaanquist:
This introduces a fix to the bug mentioned in issue #2581 by using wevtutil's archive functionality to preserve the display information of the container's OS.


freddydk:
Thanks

PR #2588, opened 2022-07-13
BCSaaS Environment Upgrade Handling

JulianSchmidtkeGWS:
Hi Freddy,

we would love to integrate more BC Admin Center API Requests into BCContainerHelper.

In this PR i start with Environment Upgrade Handling. You can check if a upgrade for an environment is scheduled by running Get-BcScheduledUpgrade and reschedule upgrades by running Reschedule-BcUpgrade.

The Functions use this API: https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/administration/administration-center-api_reschedule_updates

I tried to copy the style of your other bcsaas functions :) 

Feedback is welcome!

### DIFF
--- a/BcContainerHelper.psd1
+++ b/BcContainerHelper.psd1
@@ -135,7 +135,8 @@ FunctionsToExport = 'Add-FontsToBcContainer', 'Add-GitToAlProjectFolder',
                'Sync-BcContainerApp', 'Test-BcContainer', 'UnInstall-BcContainerApp', 
                'UnPublish-BcContainerApp', 'Wait-BcContainerReady', 
                'Write-BcContainerHelperWelcomeText', 'Invoke-gh', 'Invoke-git',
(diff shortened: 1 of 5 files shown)


freddydk:
Thanks for this - Looks good - could you add two lines to the releasenotes.txt with info on the two functions.
+ we also need to add the functions to the BcContainerHelper.psd1 file.
The .psd1 file gets generated on publish, but there are also cases, where the module is imported by running the .ps1 file and for that purpose, the .psd1 file needs to be updated as well.

JulianSchmidtkeGWS:
Hey Freddy,
i updated the releasenotes and the .psd1 file. Hope the version is correct like this

freddydk:
Thanks
Sorry for not being specific, the ReleaseNotes should be under 3.0.12 (which is the version currently in preview).

JulianSchmidtkeGWS:
Rebased and fixed the releasenotes

freddydk:
Thanks

PR #2655, opened 2022-08-29
Add extra parameters to Resolve-DependenciesFromAzureFeed

Arthurvdv:
Using the Resolve-DependenciesFromAzureFeed in our Azure DevOps pipelines I would like to expand this function with two extra parameters:
- ignoreAppVersion
- ignoredPublishers 

**ignoreAppVersion**
Would like to retrieve the most up-to-date version from Azure Feed and ignore the specific version from the app.json

**ignoredPublishers**
Some Apps aren't in our Azure Feed. Based on the Publisher would like to exclude these from trying to download.

### DIFF
--- a/PackageHandling/Resolve-DependenciesFromAzureFeed.ps1
+++ b/PackageHandling/Resolve-DependenciesFromAzureFeed.ps1
@@ -18,7 +18,13 @@
   Folder where all dependencies will be copied to
  .Parameter lvl
   Level to track recursion depth.
-    
+ .Parameter ignoredDependencies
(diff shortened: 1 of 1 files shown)


jonaswre:
@Arthurvdv looks good. I never thought about ignoring publisher but thats a very good idea.

jonaswre:
In my latest PR I used a diffrent technique for filtering dependencies. That might be a better solution here as well. I pass a filter thats then used by Where-Object what why you can basicly filter on any attribute thats in json.

PR #2714, opened 2022-09-29
Added api.partner.microsoft.com scope

jonaswre:
Maybe I'm missing a curtial part of the documentation. 
But I didn't find any reference how to setup the authentication with app source.

There were 2 issues I had. 

1. The scope to use for the authentication
2. The way to authorize an app registration against the partner center. 

This change helps with 1. addtionally it might be helpful to mention it here as well https://github.com/microsoft/AL-Go/blob/main/Scenarios/settings.md#appsource-specific-advanced-settings

If you tell me where to document the 2. issue I will do that.

### DIFF
--- a/AL-Go/New-ALGoAppSourceContext.ps1
+++ b/AL-Go/New-ALGoAppSourceContext.ps1
@@ -17,7 +17,7 @@ function New-ALGoAppSourceContext {
         $appSourceContext = [ordered]@{
             "refreshToken" = $authContext.RefreshToken
             "tenantID" = $authContext.tenantID
-            "scopes" = @("https://api.partner.microsoft.com/.default")
+            "scopes" = $authContext.scopes
         }
     }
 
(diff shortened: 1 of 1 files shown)


jonaswre:
Ok... I guess I was missing the right information.
Is there a rss feed for your blog?

freddydk:
my blog is wordpress - I think https://freddysblog.com/feed/ would work.
Based on this error - I think I will also test whether scopes is correct in New-AlGoAppSourceContext - and give an error if not.

jonaswre:
Thanks the feed is great. 

Wouldn't it be better to throw that error before the actual request? Or maybe do the request and show a warning that the scope might be wrong if you get an 401 response. Because the way you describe it in the AppSourceContext Setting I think lots of people will just write the context manually. (that's what I did at first.)

freddydk:
You mean adding a check to the invoke-IngestionAPIRestMethod function - good idea.

PR #2719, opened 2022-09-30
Add Import-BcConfigurationPackage function

sergeyol:
Add function to upload configuration package file, import and apply


sergeyol:
> 

@microsoft-github-policy-service agree [company="SMART business LLC"]

sergeyol:
@microsoft-github-policy-service agree company="SMART business LLC"

eclipses:
@sergeyol I think there is already a function that imports and applies configuration packages. 
Please have a look in the folder 
**ConfigPackageHandling**/UploadImportAndApply-ConfigPackageInBcContainer.ps1

sergeyol:
@eclipses how can it be used with online environments?

eclipses:
@sergeyol that function I pointed out works only for Nav/Bc Containers. 
Sorry, I've now realized you suggested a function that works in SaaS environments not Containers.

BcSaaS/Import-BcConfigurationPackage.ps1
```diff
@@ -0,0 +1,161 @@
...
+function Import-BcConfigurationPackage {
+    Param(
+        [Parameter(Mandatory=$true)]
+        [Hashtable] $bcAuthContext,
+        [Parameter(Mandatory=$true)]
+        [string] $environment,
+        [Parameter(Mandatory=$false)]
+        [string] $companyName,
+        [Parameter(Mandatory=$true)]
+        $packageFile,
```
freddydk, line 34: this is called configPackage in the corresponding function for containers
It would be OK to add an alias called packageFile (to both functions)

BcSaaS/Import-BcConfigurationPackage.ps1
```diff
@@ -0,0 +1,161 @@
...
+        [Parameter(Mandatory=$true)]
+        $packageCode,
+        [Parameter(Mandatory=$false)]
+        $packageName = "",
```
freddydk, line 38: packageName is auto-calculated in the other function using something like this:

```
            if (!$packageIdSpecified) {
                $packageInfo = Get-PackageInfoFromRapidStartFile $configPackage
                if ($packageInfo) {
                    $packageId = $packageInfo.Code
                }
                else {
                    $packageId = [System.IO.Path]::GetFileNameWithoutExtension($configPackage)
                    $packageId = $packageId.SubString(0, [System.Math]::Min(20, $packageId.Length))
                }
             }
```
and then the packageId is used as both name and code.
Is this wrong?

BcSaaS/Import-BcConfigurationPackage.ps1
```diff
@@ -0,0 +1,161 @@
+﻿<# 
+ .Synopsis
```
freddydk, line 2: Please move to the SaaS folder instead of BcSaaS (I might have renamed the folder)
Also please rename to Import-ConfigPackageToBcEnvironment

freddydk (changes requested):
Also, please add to releasenotes.txt and add to Bc.SaaSHelper.ps1

freddydk:
Assuming that this is no longer needed as no activity has been on this PR

Issue #2713, opened 2022-09-29
Using container with Traefik result in "404 page not found" error


DBiernat:
**Describe the issue**
I'm getting a "page not found" error message by calling the web client (https://{host}.{fqdn}/{containerName}).
Calling the web client on the host (https://{containerName}/{containerName}) is working.
Traefik Dashbord (host.ipAddress:8080/dashboard/) is reachable.

I dont't have a clue, where the problem is.

**Scripts used to create container and cause the issue**
```
New-BcContainer -accept_eula `
    -containerName $containerName `
    -artifactUrl $artifactURL `
    -vsixFile (Get-LatestAlLanguageExtensionUrl) `
    -auth UserPassword `
    -Credential $credential `
    -shortcuts None `
    -isolation hyperv `
    -licenseFile $licenseFile `
    -additionalParameters $additionalParameters `
    -myScripts $myscripts `
    -useSSL `
    -useTraefik `
    -PublicDnsName $containerPublicDns
```

**Full output of scripts**
```
BcContainerHelper is version 4.0.4
BcContainerHelper is running as administrator
Hyper-V is Enabled
UsePsSession is True
Host is Microsoft Windows Server 2022 Standard - ltsc2022
Docker Client Version is 20.10.9
Docker Server Version is 20.10.9
Removing Desktop shortcuts
Fetching all docker images
Fetching all docker volumes
Enabling SSL as otherwise all clients will see mixed HTTP / HTTPS request, which will cause problems e.g. on the mobile and modern windows clients
Using image mcr.microsoft.com/businesscentral:10.0.20348.887
PublicDnsName is {host}.{fqdn}
Creating Container TEST
Style: onprem
Multitenant: No
Version: 19.6.38118.0
Platform: 19.0.38062.0
Generic Tag: 1.0.2.13
Container OS Version: 10.0.20348.887 (ltsc2022)
Host OS Version: 10.0.20348.946 (ltsc2022)
Using hyperv isolation
Using locale de-DE
Adding special CheckHealth.ps1 to enable Traefik support
Disabling the standard eventlog dump to container log every 2 seconds (use -dumpEventLog to enable)
Downloading C:\ProgramData\BcContainerHelper\Extensions\TEST\Microsoft.VisualStudio.Services.VSIXPackage.vsix
Using license file C:\bccontainers\TEST\Lic.flf
Additional Parameters:
--volume "C:\bccontainers\TEST:C:\Run\my\bccontainer"
[...]

PR #2725, opened 2022-10-07
use traefik image with 2022 support

tfenster:
As reported in #2713, there seems to be an issue at least in some setups on Win Server 2022 to access the docker engine npipe from hyperv containers. As a fix, this uses a traefik image with a variant for Win Server 2022 so that process isolation can be used


freddydk:
Thanks Tobias

Issue #2574, opened 2022-06-30
[Question] How to Run-Al-Pipeline with a complex App Dependency Structure?


matthias-planzer:
PLEASE DO NOT INCLUDE ANY PASSWORDS OR TOKENS IN YOUR ISSUE!!!

**Describe the issue**
We have a project with the following dependency structure:
* AppDependency A: Various pre-compiled Apps
* CompiledApp A: A few Apps that should be compiled with the pipeline
  * Dependency: AppDependency A
* AppDependency B: Various pre-compiled Apps
  * Dependency: Compiled App A + AppDependencyA
* CompiledApp B: A App that should be compiled by the pipeline.
  * Dependency: AppDependency A, AppDependency B, Compiled Apps A

Now according to my knowledge, only the Steps AppDependency A + CompiledApp A are possible.
Is there a way how to do the next steps, if not, is it intended that all four steps will be possible in the future?

**Scripts used to create container and cause the issue**
```
None
```

**Full output of scripts**
```
None
```

**Screenshots**

**Additional context**

jonaswre:
Hi, are you on DevOps or GitHub? 

There is a Command Resolve-DependenciesFromAzureFeed if you use that you can handle any dependency forest, even very nested ones. 

There are only two distinctions. .app files that are stored in the feed or a multi app repo so App A and B compiled together.

I've been using that for a year or so, with almost 100 apps  and a max depth of around 10.

freddydk:
The example above isn't supported in one run with Run-AlPipeline (unless appdependenciesB and compiledappsB is test apps (then you can use TestAppDependencies and testApps)
If you want to use Run-AlPipeline, it sounds like you would have to run it twice, with the output of the first run as dependencies for the second run,

jonaswre:
Ahh yes I miss understood the question. But this should work in one compilation if you put A and B in the same Repo. 

Run-Pipeline should execute boths builds sequentially and you could provide the pre compiled apps ether using Resolve-DependenciesFromAzureFeed or just as a ZIP in the installApps property.

freddydk:
@jonaswre - you are 100% right in what you say, but as I understand the problem - F-H has a dependency on A as well.
Currently the logic in Run-AlPipeline will install all apps in InstallApps before all compiled apps.
I guess this logic could be changed to support the scenario at hand, but currently I don't believe you can.

jonaswre:
Ahhhh okay... 
[...]

PR #2766, opened 2022-11-06
Dependency graph

jonaswre:
I need an up to date master. And because the decision on #2646 is still open I had create a new PR. Sorry for that... 

I finally had the time to integrate the functions I was speaking of in https://github.com/microsoft/navcontainerhelper/issues/2574.

The Get-DependencyGraph creates a dependency graph that can be used to query dependencies and dependents for diffrent appIds in diffrent versions.

Examples are provided in Tests\DependencyGraph.Tests.ps1


jonaswre:
Maybe this should be part of the new bchelper module? Only dependency here is Get-BcContainerAppInfo

freddydk:
Probably a good idea

freddydk:
I think we should take this function and the other AzureFeed functions and place in a module called BC.AzureFeed or something like that. For now, you can have a dependency on BcContainerHelper - and when I am done splitting up containerhelper in smalller modules, you can change the dependency to the actually needed modules.

Make sense?

jonaswre:
And regarding Azure Feed I think it should be deprecated. I don't see any reason to use if you can use nuget. The way dependencies are resolved is pretty slow because there is no easily readable metadata. You need to download the app and extract it until you are able to resolve the next. Plus the disadvantage that you can't use the fourth digit of your version number. 

Migration should be pretty easy, just download all packages from azure feed and upload them to nuget. By the way it would be cool to have an option to batch upload apps to a nuget feed (Github Packages). I've created an action that takes a url and uploads it to github packages but since you can't create nuget packages from runtimes it pretty useless for uploading thirdparty apps.

jonaswre:
@freddydk I think this is still needed? 

I found it useful for walking the whole dependency tree. But if you think this is not need I would close the PR.

freddydk:
I kept the PR in case we needed the functionality in other BCCH functions, but I don't think we do.
All accepted PRs comes with a support tax - probably not a lot, but still I would need to know the details of this code.
You could create another Gallery module called BC.DependencyGraph or like - if you would like to ship this.
Thanks for the contribution though

PR #2788, opened 2022-11-17
Add extra parameters to New-NavImage

niku-live:
Using the New-NavImage/New-BcImage in our Azure DevOps pipelines I would like to expand this function with a new parameter
-   databaseBackupPath: allows selecting custom backup in place of original backup from artifacts

In our case this will save a lot of time in pipelines, because usually we are using custom database that is based on Cronus but contains all of our library apps and config. Now we need to republish all those apps for every build that depends on our library apps (even so our library apps do not change so often), so every build runs 2-5 min. longer than it could. It might not look so much of a time, but we are using pipelines a lot so it easily accumulates to hours of wait time.

With this parameter we could support scenario like this:
- Every month we run pipeline to rebuild our backups and create base build images for our target BC versions
- Every normal build pipeline uses generated base build images in PR validations and automated testing

### DIFF
--- a/ContainerHandling/New-NavImage.ps1
+++ b/ContainerHandling/New-NavImage.ps1
@@ -10,6 +10,8 @@
   Name of the image getting build. Default is myimage:<tag describing version>.
  .Parameter baseImage
   BaseImage to use. Default is using Get-BestGenericImage to get the best generic image to use.
+ .Parameter databaseBackupPath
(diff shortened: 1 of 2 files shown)


freddydk:
Could you add a line describing this in releasenotes.txt (under 4.0.7), then I can accept this.

niku-live:
> Could you add a line describing this in releasenotes.txt (under 4.0.7), then I can accept this.

Sure. Here is it is.

Since 4.0.7 release has more changes that adds new parameters to existing functions, I added my description line right after other similar changes.

PR #2791, opened 2022-11-21
Allow LCGTranslationFile compiler feature 

aholstrup1:


mazhelez (approved):
All good on my side.
The `LcgTranslationFile` feature isn't explicitly available in the compiler, it is only hidden.

However, for MS projects that use BCContainerHelper, the feature is needed to generate LCG files (used by MS for translations).

PR #2815, opened 2022-12-07
Add ability to pass compiler features through Run-ALPipeline

aholstrup1:
Add ability to pass compiler features through Run-ALPipeline

This could be useful when building an app in different modes i.e. we might want a different set of compiler features when building the translated app vs. in CLEAN mode


freddydk:
I forgot to bump the version number - just did that.
Please pull last changes and update the releasenotes.txt with a line about the added parameter.

Thanks

Issue #2853, opened 2023-01-13
Missing cu version in nav2018 Artifact Manifest


ChrisBlankDe:
**Describe the issue**
Some of the newest NAV 2018 Artifact (starting with cu53) are missing the cu version in the artifact manifest.json.

**Scripts used to reproduce**
```posh
Get-NavArtifactUrl -nav 2018 -cu 49 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu #ok
Get-NavArtifactUrl -nav 2018 -cu 50 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu #ok
Get-NavArtifactUrl -nav 2018 -cu 51 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu #ok
Get-NavArtifactUrl -nav 2018 -cu 52 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu #ok
Get-NavArtifactUrl -nav 2018 -cu 53 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu
Get-NavArtifactUrl -nav 2018 -cu 54 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu
Get-NavArtifactUrl -nav 2018 -cu 55 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu
Get-NavArtifactUrl -nav 2018 -cu 56 -country w1 | %{Download-Artifacts -artifactUrl $_} | Join-Path -ChildPath 'manifest.json'| gi| gc | ConvertFrom-Json | select Version,cu
[...]

PR #2856, opened 2023-01-13
Fix Missing cu version in nav2018 Artifact Manifest

ChrisBlankDe:
Fix for #2853


freddydk:
Please add a line in the top section of releasenotes that the issue is resolved and a small text

PR #2952, opened 2023-03-27
Added dependencyPublishingOption param to Publish-BcContainerApp function

left-code:
When republishing parent app with dependencies using `Publish-BcContainerApp` with option `-useDevEndpoint` request returns **error 422**:

```
Status Code 422 : Unprocessable Entity
The extension <CHILD EXTENSION NAME> <VERSION> cannot be installed because a newer version <CURRENT VERSION> is already installed.
```
Parent app is published and installed, but all dependent apps are uninstalled. 

If `DependencyPublishingOption` is set to **default** error is not returned and dependencies are not uninstalled.

Therefore I suggest to add new parameter to function.

### DIFF
--- a/AppHandling/Publish-NavContainerApp.ps1
+++ b/AppHandling/Publish-NavContainerApp.ps1
@@ -242,10 +242,7 @@ try {
                 elseif ($syncMode -eq "ForceSync") {
                     $schemaUpdateMode = "forcesync"
                 }
-                $url = "$devServerUrl/dev/apps?SchemaUpdateMode=$schemaUpdateMode"
-                if ($PSBoundParameters.ContainsKey('DependencyPublishingOption')) {
-                    $url += "&DependencyPublishingOption=$dependencyPublishingOption"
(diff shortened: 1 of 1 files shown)


freddydk:
The code needs to be changed so that if the parameter is not specified, then it is not added to the URL.
Earlier versions of BC will fail if this parameter is added.

left-code:
As I can see it was hardcoded to **ignore**, and I set default value to **ignore** as well.
I would say if it wasn't added it should act like before. Do I miss something?

freddydk:
if ($PSBoundParameters.ContainsKey('DependencyPublishingOption')) {
  # add value to URL
}

freddydk:
Sorry, my bad - the parameter was always there, your first commit was correct.
I changed it back and will merge the PR

Thanks

freddydk:
Will be in the next preview in a few hours.

freddydk:
Should be in preview now.

Issue #2963, opened 2023-04-04
Update Get-LatestAlLanguageExtensionUrl to support Pre-Release versions


jwikman:
**Describe the issue**
Microsoft has now released a Pre-Release of AL Language Extension (see https://marketplace.visualstudio.com/items?itemName=ms-dynamics-smb.al)

Get-LatestAlLanguageExtensionUrl does not support filtering away those vsix, so right now it'll always return the Pre-Release version (the latest version).

**Scripts used to create container and cause the issue**
```
Get-LatestAlLanguageExtensionUrl
```

**Full output of scripts**
```
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/12.0.779795/1680613254602/Microsoft.VisualStudio.Services.VSIXPackage
```
**Expected output**
```
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/11.0.779295/1680528043617/Microsoft.VisualStudio.Services.VSIXPackage
```

**Solution**
I'll submit a PR that solves this, it could be used as-is - or as an inspiration for your own solution.

jwikman:
Solved by #2965

PR #2964, opened 2023-04-04
Support Pre-Release of AL Language #2963

jwikman:
This PR adds support for the Pre-Release versions of AL Language extension.

The changes looks a bit worse than they are, due to formatting changes.

- I added a new parameter that can be used if you want to download the latest Pre-Release version
- the `flags` attribute in the request are changed from 914 to 55 (that is what VSCode uses while querying the VSCode Marketplace)
- A filter is set on `Microsoft.VisualStudio.Code.PreRelease` to exclude pre-release versions if the new parameter is not specified
  - I'm using `IndexOf()` together with array syntax to handle null values of the `Microsoft.VisualStudio.Code.PreRelease` property...

### DIFF
--- a/ContainerHandling/Get-LatestAlLanguageExtensionUrl.ps1
+++ b/ContainerHandling/Get-LatestAlLanguageExtensionUrl.ps1
@@ -3,39 +3,47 @@
   Get the URL of the latest AL Language Extension from VS Code Marketplace
  .Description
   Get the URL of the latest AL Language Extension from VS Code Marketplace
+ .Parameter includePreRelease
+  Allow pre-release versions of the AL Language Extension
  .Example
   New-BcContainer ... -vsixFile (Get-LatestAlLanguageExtensionUrl) ...
  .Example
   Download-File -SourceUrl (Get-LatestAlLanguageExtensionUrl) -DestinationFile "c:\temp\al.vsix"
 #>
 function Get-LatestAlLanguageExtensionUrl {
-$telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
-try {
+    [CmdletBinding()]
+    Param (
+        [switch] $includePreRelease
+    )
 
-    $listing = Invoke-WebRequest -Method POST -UseBasicParsing `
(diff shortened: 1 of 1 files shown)


freddydk:
Hi Johannes - we came to almost the same solution.
My PR here: https://github.com/microsoft/navcontainerhelper/pull/2965

The flags are described here:
enum Flags {
/**
* None is used to retrieve only the basic extension details.
*/
None = 0x0,

/**
* IncludeVersions will return version information for extensions returned
*/
IncludeVersions = 0x1,

/**
* IncludeFiles will return information about which files were found
* within the extension that were stored independent of the manifest.
* When asking for files, versions will be included as well since files
* are returned as a property of the versions.
* These files can be retrieved using the path to the file without
* requiring the entire manifest be downloaded.
*/
IncludeFiles = 0x2,

/**
* Include the Categories and Tags that were added to the extension definition.
*/
IncludeCategoryAndTags = 0x4,

/**
* Include the details about which accounts the extension has been shared
* with if the extension is a private extension.
*/
IncludeSharedAccounts = 0x8,

/**
* Include properties associated with versions of the extension
*/
IncludeVersionProperties = 0x10,

/**
* Excluding non-validated extensions will remove any extension versions that
* either are in the process of being validated or have failed validation.
*/
ExcludeNonValidated = 0x20,

/**
* Include the set of installation targets the extension has requested.
*/
IncludeInstallationTargets = 0x40,

/**
* Include the base uri for assets of this extension
*/
IncludeAssetUri = 0x80,

/**
* Include the statistics associated with this extension
*/
IncludeStatistics = 0x100,

/**
* When retrieving versions from a query, only include the latest
* version of the extensions that matched. This is useful when the
* caller doesn't need all the published versions. It will save a
* significant size in the returned payload.
*/
IncludeLatestVersionOnly = 0x200,

/**
* This flag switches the asset uri to use GetAssetByName instead of CDN
* When this is used, values of base asset uri and base asset uri fallback are switched
* When this is used, source of asset files are pointed to Gallery service always even if CDN is available
*/
Unpublished = 0x1000,

/**
* Include the details if an extension is in conflict list or not
*/
IncludeNameConflictInfo = 0x8000,
}

jwikman:
Ok, good. I'll abandon this then. 👍

PR #3020, opened 2023-04-28
New functions "Copy-ItemFromBcContainer" & "Copy-ItemToBcContainer"

hemisphera:
I quite often need to copy entire folder trees from/to a container. The cmdlets `Copy-File[To/From]BcContainer` however do not allow this.

This proposal implements two new cmdlets called `Copy-Item[To/From]BcContainer`. I chose to create new cmdlets and **not** change the existing ones, because behavior might be slightly different. I tried to keep it consistent, though:
- Copying files should behave the same as the `File` pendant
- Copying folders will copy all folder contents to `containerPath`. If the folder does not exist, it is created. The _contents_ of the source folder are then copied to the target folder. This is the same for both **from** and **to**.


Misc/Copy-ItemFromBcContainer.ps1
```diff
@@ -0,0 +1,66 @@
...
+  Copy an item (file or folder) from a container to the host system.
+ .Parameter containerName
+  Name of the container from which you want to copy the item.
+ .Parameter containerPath
+  Path to the item in the container which has to be copied.
+ .Parameter localPath
+  Path on the host where the item will be placed.
+  If the source item is a folder, 'localPath' will be created (or treated) as a folder and all items inside the source folder will be copied to it.
+ .Example
+  Copy-ItemFromBcContainer -containerName test2 -containerPath "c:\run\my" -localPath "c:\temp\copy-of-my"
```
freddydk, line 14: The sample is not good as c:\run\my inside the container is already shared.
Maybe grab c:\build to c:\temp\build

Misc/Copy-ItemFromBcContainer.ps1
```diff
@@ -0,0 +1,66 @@
...
+#>
+function Copy-ItemFromBcContainer {
+    Param (
+        [string] $containerName = $bcContainerHelperConfig.defaultContainerName,
+        [Parameter(Mandatory = $true)]
+        [string] $containerPath,
+        [Parameter(Mandatory = $false)]
+        [string] $localPath = $containerPath
```
freddydk, line 22: localPath should be mandatory true and not defaulted

Misc/Copy-ItemFromBcContainer.ps1
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
+    try {
+
+        if (!(Test-BcContainer -containerName $containerName)) {
+            throw "Container $containerName does not exist"
+        }
+        Write-Host "Copy '$containerPath' from container $containerName to '$localPath' on host"
```
freddydk, line 31: I think we should test whether localpath exists here and throw early

Misc/Copy-ItemToBcContainer.ps1
```diff
@@ -0,0 +1,66 @@
...
+ .Example
+  Copy-ItemToBcContainer -containerName test2 -localPath "c:\temp\myfile.txt" -containerPath "c:\run\my\myfile.txt"
+#>
+function Copy-ItemToBcContainer {
+    Param (
+        [string] $containerName = $bcContainerHelperConfig.defaultContainerName,
+        [Parameter(Mandatory = $true)]
+        [string] $localPath,
+        [Parameter(Mandatory = $false)]
+        [string] $containerPath = $localPath
```
freddydk, line 22: when copying into the container, I think it is OK to default containerPath - you cannot really overwrite anything dangerous.

Misc/Copy-ItemToBcContainer.ps1
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
+    try {
+
+        if (!(Test-BcContainer -containerName $containerName)) {
+            throw "Container $containerName does not exist"
+        }
+        Write-Host "Copy '$localPath' from host to '$containerPath' on container $containerName"
```
freddydk, line 31: also test here whether containerPath exists - if it does, fail early

freddydk (changes requested):

Misc/Copy-ItemFromBcContainer.ps1
hemisphera: While I do agree, I mimicked what is done in `Copy-FileFromBcContainer`. But maybe there it's purely for backward compatibility? Anyways ... will fix.

Misc/Copy-ItemFromBcContainer.ps1
hemisphera: The idea is that local path actually _can_ exist. If it does, the contents from source are added _inside_ target. I am still not sure whether that's a good idea or failing is better.

Misc/Copy-ItemToBcContainer.ps1
hemisphera: Same as in `Copy-ItemFromBcContainer` more or less. But I'd rather not fail at all here. I imagine a scenario where I stage multiple files from the host into the container with multiple calls. Failing would force me to stage on the host first and then copy to the container with one single call.

Misc/Copy-ItemFromBcContainer.ps1
hemisphera: Fixed.

Misc/Copy-ItemFromBcContainer.ps1
freddydk: Ahh, I didn't notice that CreateDirectory only creates the directory if it doesn't already exist.

Misc/Copy-ItemToBcContainer.ps1
freddydk: Yeah, just stating that you don't need to make containerPath mandatory here.
The only reason for doing this on the other command would be that the function could mess up your local machine pretty bad by transferring one wrong parameter.

Misc/Copy-ItemToBcContainer.ps1
freddydk: got it and agree

freddydk (approved):

PR #3035, opened 2023-05-04
Allow 3 chars for all analyzer rules i.e. PTE1234

MODUSCarstenScholling:
Hi Freddy,

This very small PR adds the ability to parse custom 3 character issue codes from alc output. I tried to find a gap for PTE, but without luck. Please see this as a compatibility PR for custom analyzers :).

Some years back, we decided to implement MCP which is not fully covered by the current implementation. 

Thank you,
Carsten


MODUSCarstenScholling:
Thank you!

PR #3046, opened 2023-05-11
Allow Downloading Core Artifact

aholstrup1:
You would then be able to download the core artifact like: 
Download-Artifacts -artifactUrl (Get-BCArtifactUrl -country 'core') 

Right now this fails when looking for the manifest.json

### DIFF
--- a/Artifacts/Download-Artifacts.ps1
+++ b/Artifacts/Download-Artifacts.ps1
@@ -129,36 +129,39 @@ try {
                 }
             }
             try { [System.IO.File]::WriteAllText((Join-Path $appArtifactPath 'lastused'), "$([datetime]::UtcNow.Ticks)") } catch {}
-    
-            $appManifestPath = Join-Path $appArtifactPath "manifest.json"
-            $appManifest = Get-Content $appManifestPath | ConvertFrom-Json
 
-            # Patch wrong license file in ONPREM AU version 20.5.45456.45889
-            if ($artifactUrl -like '*/onprem/20.5.45456.45889/au') {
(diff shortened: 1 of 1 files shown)


freddydk:
so, you would use 

$coreUrl = Get-BCArtifactUrl -version 22 -country core

to get the URL?

Artifacts/Download-Artifacts.ps1
```diff
@@ -129,36 +129,39 @@ try {
...
-                $appManifest | ConvertTo-Json | Set-Content -Path $appManifestPath
-            }
-    
-            if ($appManifest.PSObject.Properties.name -eq "applicationUrl") {
-                $redir = $true
-                $artifactUrl = $appManifest.ApplicationUrl
-                if ($artifactUrl -notlike 'https://*') {
-                    $artifactUrl = "https://$($appUri.Host)/$artifactUrl$($appUri.Query)"
+            $coreArtifact = $appUri.AbsolutePath -like "*/core"
+            if (-not $coreArtifact) {
```
freddydk, line 134: Instead of checking for the name - just check whether appManifestPath exists.
This code section is only for patching wrong app manifests shipped - and should probably have an if (Test-Path around it).
This would also allow me to download a platform without the app part if needed.

Artifacts/Download-Artifacts.ps1
```diff
@@ -129,36 +129,39 @@ try {
                 }
```
freddydk, line 129: In the above section it mentions downloading application artifact and unpacking application artifact.
Remove the word "application" as it now can be core or platform

freddydk (changes requested):

freddydk (approved):

Issue #3051, opened 2023-05-16
Unable to import .rapidstart file - The property 'Code' cannot be found on this object


Drakonian:
**Describe the issue**
BCCContainerHelper 5.0.2
The Get-PackageInfoFromRapidStartFile function contains a bug that makes it impossible to import a Package via the Run-AlPipeline function. 
I am attaching a pull request with the fix. Even if you see more correct fix at least you know problem place :)

**Scripts used to create container and cause the issue**
```
UploadImportAndApply-ConfigPackageInBcContainer
```

**Full output of scripts**
```
UploadImportAndApply-ConfigPackageInBcContainer Telemetry Correlation Id: 46be8886-fedb-4de8-a978-7ee1ee78bf6d
##[error]The property 'Code' cannot be found on this object. Verify that the property exists.
At C:\Program Files\WindowsPowerShell\Modules\BcContainerHelperUK\5.0.2\ConfigPackageHandling\UploadImportAndApply-Conf
igPackageInBcContainer.ps1:46 char:21
+                     $packageId = $packageInfo.Code
+                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], PropertyNotFoundException
    + FullyQualifiedErrorId : PropertyNotFoundStrict
##[error]PowerShell exited with code '1'.

...
```

freddydk:
Thanks for the help, much appreciated.

freddydk:
Shipped in 5.0.3

PR #3052, opened 2023-05-16
Issue #3051 Unable to import .rapidstart file - The property 'Code' cannot be found on this object

Drakonian:
Issue #3051 Unable to import .rapidstart file - The property 'Code' cannot be found on this object


freddydk:
Looks good, thanks.
Either that - or replace the break with exit I guess.

Issue #3066, opened 2023-05-23
Disable-SSL hook does not work with PS7


warlof:
**Describe the issue**

It appears the following custom hook which normally allow to bypass ssl certificate validation does not work with Powershell 7 environment :

```
        $sslVerificationDisabled = ($protocol -eq "https://")
        if ($sslVerificationDisabled) {
            if (-not ([System.Management.Automation.PSTypeName]"SslVerification").Type)
            {
                Add-Type -TypeDefinition "
                    using System.Net.Security;
                    using System.Security.Cryptography.X509Certificates;
                    public static class SslVerification
                    {
                        private static bool ValidationCallback(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) { return true; }
                        public static void Disable() { System.Net.ServicePointManager.ServerCertificateValidationCallback = ValidationCallback; }
                        public static void Enable()  { System.Net.ServicePointManager.ServerCertificateValidationCallback = null; }
                    }"
            }
            Write-Host "Disabling SSL Verification"
            [SslVerification]::Disable()
        }
```

For a reason I cannot understand, the HttpClient simply do not care about it.

I detected this while using the following command in order to build an app : `Compile-AppInNavContainer.ps1`
When running it with Powershell 7, I get the following output :

```
Downloading using HttpClient
ERROR Exception calling "GetResult" with "0" argument(s): "The SSL connection could not be established, see inner exception."
Compile-AppInBCContainer Telemetry Correlation Id: c87d90ca-7f37-490a-87f4-e18b247026ed
Exception calling "GetResult" with "0" argument(s): "The SSL connection could not be established, see inner exception."
At C:\Program Files\WindowsPowerShell\Modules\BcContainerHelper\5.0.3\AppHandling\Compile-AppInNavContainer.ps1:488 char:25
+                         throw (GetExtendedErrorMessage $_)
+                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : OperationStopped: (Exception calling "…e inner exception.":String) [], RuntimeException
+ FullyQualifiedErrorId : Exception calling "GetResult" with "0" argument(s): "The SSL connection could not be established, see inner > exception."
##[error]PowerShell exited with code '1'.
```

However, if it's run with Powershell core or 5.1, everything is working without any issue.

[...]

PR #3067, opened 2023-05-23
fix: enhance ps7 compatibility with ssl selfsigned

warlof:
for some reason, it appears ssl validation bypass using System.Net.ServicePointManager does not work with powershell 7
use built-in powershell method Invoke-WebRequest which provide same capabilities

related to #3066

### DIFF
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -485,6 +485,14 @@ try {
                     }
                     if ($throw) {
                         Write-Host "ERROR $($_.Exception.Message)"
+                        if ($_.Exception.InnerException)
+                        {
+                            Write-Host "ERROR $($_.Exception.InnerException.Message)"
+                            if ($_.Exception.InnerException.InnerException)
+                            {
+                                Write-Host "ERROR $($_.Exception.InnerException.InnerException.Message)"
+                            }
+                        }
                         throw (GetExtendedErrorMessage $_)
                     }
                 }
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -391,25 +391,8 @@ try {
             $devServerUrl = "$($protocol)$($containerName):$($customConfig.DeveloperServicesPort)/$ServerInstance"
         }
     
-        $sslVerificationDisabled = ($protocol -eq "https://")
-        if ($sslVerificationDisabled) {
-            if (-not ([System.Management.Automation.PSTypeName]"SslVerification").Type)
-            {
-                Add-Type -TypeDefinition "
-                    using System.Net.Security;
-                    using System.Security.Cryptography.X509Certificates;
-                    public static class SslVerification
-                    {
-                        private static bool ValidationCallback(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) { return true; }
-                        public static void Disable() { System.Net.ServicePointManager.ServerCertificateValidationCallback = ValidationCallback; }
-                        public static void Enable()  { System.Net.ServicePointManager.ServerCertificateValidationCallback = null; }
-                    }"
-            }
-            Write-Host "Disabling SSL Verification"
-            [SslVerification]::Disable()
-        }
-    
         $timeout = 300000
+        $sslVerificationDisabled = ($protocol -eq "https://")    
         if ($customConfig.ClientServicesCredentialType -eq "Windows") {
             $useDefaultCredentials = $true
         }
@@ -467,7 +450,7 @@ try {
                 }
                 Write-Host "Url : $Url"
                 try {
-                    DownloadFileLow -sourceUrl $url -destinationFile $symbolsFile -timeout $timeout -useDefaultCredentials:$useDefaultCredentials -Headers $headers
+                    DownloadFileLow -sourceUrl $url -destinationFile $symbolsFile -timeout $timeout -useDefaultCredentials:$useDefaultCredentials -Headers $headers -skipCertificateCheck:$sslVerificationDisabled
                 }
                 catch {
                     $throw = $true
@@ -485,14 +468,6 @@ try {
                     }
                     if ($throw) {
                         Write-Host "ERROR $($_.Exception.Message)"
-                        if ($_.Exception.InnerException)
-                        {
-                            Write-Host "ERROR $($_.Exception.InnerException.Message)"
-                            if ($_.Exception.InnerException.InnerException)
-                            {
-                                Write-Host "ERROR $($_.Exception.InnerException.InnerException.Message)"
-                            }
-                        }
                         throw (GetExtendedErrorMessage $_)
                     }
                 }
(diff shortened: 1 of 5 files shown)


freddydk:
I always found Invoke-WebRequest to be 20-30% slower than WebClient download.
Not sure whether that is also the case in PS7.
(This was the reason for using HttpClient in the first place).

I will add a few questions in the code

HelperFunctions.ps1
```diff
@@ -987,39 +987,68 @@ function DownloadFileLow {
...
-        try {
-            $stream = $httpClient.GetStreamAsync($sourceUrl).GetAwaiter().GetResult()
-            $fileStream = New-Object System.IO.Filestream($destinationFile, $fileMode)
-            $stream.CopyToAsync($fileStream).GetAwaiter().GetResult() | Out-Null
-            $fileStream.Close()
-        }
-        finally {
-            if ($fileStream) {
-                $fileStream.Dispose()
+            Write-Host "Downloading using HttpClient"
```
freddydk, line 1018: We might not need HttpClient at all - if Invoke-WebRequest is fast enough, then httpClient is probably not needed.

HelperFunctions.ps1
```diff
@@ -987,39 +987,68 @@ function DownloadFileLow {
...
-        $fileStream = $null
-        if ($dontOverwrite) {
-            $fileMode = [System.IO.FileMode]::CreateNew
+        if ($PSVersionTable.PSVersion -ge "7.0.0") {
+            Write-Host "Downloading using Invoke-WebRequest"
+            if ($useDefaultCredentials) {
+                $response = Invoke-WebRequest -UseDefaultCredentials -TimeoutSec $timeout -Headers $headers -Uri $sourceUrl -SkipCertificateCheck
+            }
+            else {
+                $response = Invoke-WebRequest -TimeoutSec $timeout -Headers $headers -Uri $sourceUrl -SkipCertificateCheck
```
freddydk, line 996: isn't -useBasicParsing needed as well?

AppHandling/Compile-AppInNavContainer.ps1
```diff
@@ -485,6 +485,14 @@ try {
                     }
                     if ($throw) {
                         Write-Host "ERROR $($_.Exception.Message)"
+                        if ($_.Exception.InnerException)
+                        {
```
freddydk, line 489: Isn't these things part of GetExtendedErrorMessage as well?

freddydk (changes requested):

freddydk:
If you prefer that I do these changes myself, let me know and I will take over.

HelperFunctions.ps1
warlof: `Invoke-WebRequest` signature has changed between PowerShell versions
`SkipCertificateCheck` Switch is not available for version prior PowerShell 7

So, in case of ps7 - we use it (first block)
Otherwise, we stick to the current implementation which is working fine anyway (except with ps7 😂)

HelperFunctions.ps1
warlof: Non longer with PowerShell 7, according to the documentation - it is native behavior since PowerShell 6

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.3#-usebasicparsing

AppHandling/Compile-AppInNavContainer.ps1
warlof: Maybe truncated ? But without this part, I didn't had the reason of failure (only the fact that there was an issue with SSL channel and the trace)

I'll double check - but with this change, it disclose up to the fact issue is related to not trusted certificate.

warlof:
Hey @freddydk , doesn't seems to be a big deal to enhance the pr with your requests. I can probably sort them this week.

Minus two things
- UseBasicParser switch is deprecated since PowerShell 6, since changes from this PR have been designed to address SSL issue with PowerShell 7 and the current implementation, I don't think we should add it
- SkipSslCertificateCheck switch is not available prior PowerShell 7

HelperFunctions.ps1
freddydk: HttpClient was actually only introduced for PS7 and later (as WebClient doesn't work either on PS7 or on Linux)

AppHandling/Compile-AppInNavContainer.ps1
warlof: Just had a quick look in it
It might be due to the level of the exception from which you get enough details when an SSL exception occurred (third one here)

As per what I understand from the current GetExtendedErrorMessage - it goes only up to the second level.

By looking on other points, I'll see if I cannot simply enhance this method instead.

HelperFunctions.ps1
```diff
@@ -987,39 +987,68 @@ function DownloadFileLow {
...
+            }
+
+            $fileStream = $null
```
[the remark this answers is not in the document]
warlof: Doing this will prevent usage of `$dontOverwrite`

AppHandling/Publish-NavContainerApp.ps1
```diff
@@ -1,3 +1,21 @@
+$sslCallbackCode = @"
+	using System.Net.Security;
+	using System.Security.Cryptography.X509Certificates;
+
+	public static class SslVerification
+	{
```
freddydk, line 6: Why would you need this here (as well as in HelperFunctions)
The variables and types from HelperFunctions should always be loaded when running.

HelperFunctions.ps1
```diff
@@ -1,1121 +1,1171 @@
...
+
+function GetExtendedErrorMessage {
+    Param(
+        $errorRecord
+    )
+
+    $exception = $errorRecord.Exception
+    $message = $exception.Message
+
+    if ($errorRecord.ErrorDetails) {
```
freddydk, line 459: This will fail if ErrorDetails isn't a property, please move it inside the try / catch

freddydk (changes requested):

freddydk:
I really like the changes here, very nice.
Two small questions/change requests

freddydk:
> So, I restore initial codebase and adapt existing part of HttpClient usage :
> 
> **About initial SSL issue with PS7**
> 
> HttpClient does not relay on ServicePointManager to validate SSL certificates I extend the existing SSL hook class by renaming and changing visibility of the certificate validation callback
> 
> As a result, it is usable by both http client implementation
> 
> * when `WebClient`, we call `Disable` or `Enable` static methods
> * when `HttpClient`, we bind the class callback directly
> 
> **About refactor of existing SSL hook class**
> 
> I introduced a `skipCertificateCheck` switch into method `DownloadFileLow` which allow to remove part of duplicated code When the switch is `$true`, we disable SSL Certificate check Also, in order to keep output consistent between `WebClient` and `HttpClient`, I always print state of enable/disable SSL check
> 
> We might still enhance it a little bit by loading in a single place - but I'm not fluent enough with powershell to know where and how - so we could get rid of all `[SslVerification]::Disable()` and `[SslVerification]::Enable()` duplication :)
> 
> * https://github.com/microsoft/navcontainerhelper/blob/b495deb26077badc2ae811543967cbe1da8dba03/Api/Invoke-NavContainerApi.ps1#L118-L129
> * https://github.com/microsoft/navcontainerhelper/blob/b495deb26077badc2ae811543967cbe1da8dba03/AppHandling/PsTestFunctions.ps1#L1244-L1256
> 
> **About Error Handling**
> 
> The `GetExtendedErrorMessage` was based on `WebClient` - however, `HttpClient` does not throw same `Exception` object. As a result, the handler was failing when it attempt to cast an `HttpClientException` into a `WebException` Therefore, I added a type check on the second and third level `InnerException` and print the relevant message which lead to more verbose error handling.
> 
> Last but not least, depending on the Exception, the `ErrorDetail` property might be null which crash the overall script. I therefore added an extra check at this stage.

I really like the changes here, very nice.
Two small questions/change requests

AppHandling/Publish-NavContainerApp.ps1
warlof: then it is indeed unneeded - sorted :)

warlof:
both feedbacks have been handled with last commit, thank you for the review :)

freddydk (approved):

Issue #3075, opened 2023-05-31
Install/Upgrade Azure CLI not working on Install-AzDevop


Arthurvdv:
**Describe the issue**
I'm having troubles to get the Install-AzDevop working on both (install and upgrade) scenario's. I'll start on the upgrade scenario first.

**B) Upgrade**
When running the Install-AzDevop on a machine where Azure CLI is installed, it isn't upgraded.

**Scripts used to create container and cause the issue**
```powershell
Import-Module BcContainerHelper -DisableNameChecking
Install-AzDevops

$az_upgrade = Write-Output n | az upgrade 2>&1
$az_upgrade | Select-String "Latest version available"
Write-Host $az_upgrade
```

**Full output of scripts**
```
Import-Module BcContainerHelper -DisableNameChecking
Install-AzDevops

$az_upgrade = Write-Output n | az upgrade 2>&1
$az_upgrade | Select-String "Latest version available"
Write-Host $az_upgrade

WARNING: Your current Azure CLI version is 2.42.0. Latest version available is 2.49.0.
WARNING: This command is in preview and under development. Reference and support levels: https://aka.ms/CLI_refstatus WARNING: Your current Azure CL
I version is 2.42.0. Latest version available is 2.49.0. WARNING: Please check the release notes first: https://docs.microsoft.com/cli/azure/release
-notes-azure-cli ERROR: No tty available. Please run command with --yes.
```

I would expect the version to be upgraded from 2.42.0 to 2.49.0.

**A) Install**
When running the Install-AzDevop on a machine where Azure CLI isn't installed it throws an error.

**Scripts used to create container and cause the issue**
```powershell
Import-Module BcContainerHelper -DisableNameChecking
Install-AzDevops
```

**Full output of scripts**
```
Install-AzDevops Telemetry Correlation Id: f380d4b0-8936-4162-ae4b-a2b44a2daab1
az : The term 'az' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a 
path was included, verify that the path is correct and try again.
At C:\Program Files\WindowsPowerShell\Modules\bccontainerhelper\5.0.3\PackageHandling\Install-AzDevops.ps1:23 char:19
+     $extensions = az extension list -o json | ConvertFrom-Json
+                   ~~
    + CategoryInfo          : ObjectNotFound: (az:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CommandNotFoundException
...
```

[...]

PR #3076, opened 2023-06-01
Fixed Azure CLI install 

jonaswre:
this should solve #3075

### DIFF
--- a/PackageHandling/Install-AzDevops.ps1
+++ b/PackageHandling/Install-AzDevops.ps1
@@ -10,16 +10,19 @@ function Install-AzDevops {
 
 $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
 try {
-
-    Try {
+    try {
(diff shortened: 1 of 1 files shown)


PackageHandling/Install-AzDevops.ps1
```diff
@@ -14,12 +14,14 @@ try {
     Try {
         $az_upgrade = Write-Output n | az upgrade 2>&1
```
freddydk, line 15: Why not just run 
image
and then if this fails (i.e. AZ is not installed), then do the install below=

freddydk (changes requested):

PackageHandling/Install-AzDevops.ps1
freddydk: Because that comes with a UI:-)

freddydk (approved):

PackageHandling/Install-AzDevops.ps1
jonaswre: Yes and az version doesn't tell me if there is a new version. 
This is the only trick i've found to check if an update is available.

PR #3080, opened 2023-06-05
Alpaca functions

freddydk:
+ a small fix in AL-Go Context


Alpaca/New-AlpacaBcContainer.ps1
```diff
@@ -0,0 +1,146 @@
...
+    $additionalFolders = @( "c:\run=$runFolderUrl" ) + $additionalFolders
+
+    $parameters = [ordered]@{
+        "persistence" = $devContainer.IsPresent
+        "environmentVars" = $environmentVariables
+        "additionalLabels" = $additionalLabels
+        "additionalFolders" = $additionalFolders
+        "licenseFile" = $licenseFileUrl
+        "bcArtifact" = @{
+            "storageAccount" = $artifactUrl.Split('/')[2].Split('.')[0]
```
lippertmarkus, line 117: We have an "intelligent image selection" in the service that prefers images that are already available so that image builds happen very rarely. We instead rebuild images every night so that all new images are available for the next day and only rebuild during the day if the user requests a specific version that isn't available yet.

To make this "intelligent image selection" work best, we would instead need the requested version instead of the full version. Fo example if the user requests BC 20, container will be created with newest BC 20 version that is already available. Every night the newest image will be rebuilt so the user might get new images delayed by 1 day.

So the best would be if `bcArtifact` has the same values that a user would pass into `get-bcartifacturl`

lippertmarkus (changes requested):
Great work, thanks! Just some small nits.

Alpaca/New-AlpacaBcContainer.ps1
```diff
@@ -0,0 +1,146 @@
...
+        [string[]] $additionalFolders = @(),
+        [PSCredential] $credential,
+        [switch] $enableSymbolLoading,
+        [switch] $enablePremium,
+        [switch] $enablePerformanceCounter,
+        [switch] $publishSwaggerUI,
+        [string] $autoStartTime = '08:00',
+        [string] $autoStartTimeUTC = [datetime]::ParseExact($autoStartTime,'H:mm',[CultureInfo]::InvariantCulture).ToUniversalTime().ToString('HH:mm'),
+        [ValidateSet('deactivated','daily','weekdays','once')]
+        [string] $autoStartMode = "$(if($devContainer){"weekdays"}else{"deactivated"})",
```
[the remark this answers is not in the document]
freddydk: fixed

Alpaca/New-AlpacaBcContainer.ps1
freddydk: OK, I have changed this to support the format of the artifact parameter in AL-Go.
Either the full ArtifactUrl if they indeed need that specific artifact
or a string using storageAccount/type/version/country (where version can be like 21.5)

lippertmarkus: Understood. But how could I specify select=nextmajor this way?

freddydk: I actually have a select as well storageAccount/type/version/country/select
How should I do this with Alpaca?

freddydk: meaning that ///us/nextmajor would give me nextmajor (which corresponds to giving these parameters to get-bcartifactUrl)

lippertmarkus: there's another field `select` in `bcArtifact` that then can be set to the according value

freddydk: and what about the Insider sas token - is that also a field then?

lippertmarkus: that is set up in the backend per-tenant. Btw I remember that you thought about removing insider sas token requirement completely - what's the state on this? would make onboarding even easier

PR #3086, opened 2023-06-14
Add parameters for scripts to run pre and post compilation

mazhelez:
Add parameters `PreCompileApp` and `PostCompileApp` to `Run-AlPipeline`: script blocks to run before and after compiling an app when setting `useCompilerFolder` switch.

TODO:
- [x] Update release notes

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -187,6 +187,19 @@
         [ref] $compilationParams
     )
     ...
+    # Change the output folder based on the app type
+    switch($appType) {
+        "app" {
+            $compilationParams.Value.appOutputFolder = "MyApps"
+        }
+        "testApp" {
+            $compilationParams.Value.appOutputFolder = "MyTestApps"
+        }
+        "bcptApp" {
+            $compilationParams.Value.appOutputFolder = "MyBcptApps"
+        }
+    }
+    ...
   }
  .Parameter PostCompileApp
   Custom script to run after compiling an app.
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -120,6 +120,8 @@
   Include this switch to indicate that you do not want to publish the app. Including this switch will also mean that upgrade won't happen and tests won't run.
  .Parameter uninstallRemovedApps
   Include this switch to indicate that you want to uninstall apps, which are included in previousApps, but not included (upgraded) in apps, i.e. removed apps
+ .Parameter useCompilerFolder
+  Include this switch to indicate that you want to use the compiler folder instead of creating a docker container for app compilation.
  .Parameter reUseContainer
   Including the reUseContainer switch causes pipeline to reuse the container with the given name if it exists
  .Parameter keepContainer
@@ -174,6 +176,10 @@
   Override function parameter for Import-TestToolkitToBcContainer
  .Parameter CompileAppInBcContainer
   Override function parameter for Compile-AppInBcContainer
+ .Parameter PreCompileApp
+  Custom script to run before compiling an app. The script will be run only if `useCompilerFolder` switch is set.
+ .Parameter PostCompileApp
+  Custom script to run after compiling an app. The script will be run only if `useCompilerFolder` switch is set.
  .Parameter GetBcContainerAppInfo
   Override function parameter for Get-BcContainerAppInfo
  .Parameter PublishBcContainerApp
@@ -274,6 +280,7 @@ Param(
     [switch] $doNotPerformUpgrade,
     [switch] $doNotPublishApps,
     [switch] $uninstallRemovedApps,
+    [Parameter(ParameterSetName='UseCompilerFolder')]
     [switch] $useCompilerFolder = $bcContainerHelperConfig.useCompilerFolder,
     [switch] $reUseContainer,
     [switch] $keepContainer,
@@ -301,6 +308,10 @@ Param(
     [scriptblock] $SetBcContainerKeyVaultAadAppAndCertificate,
     [scriptblock] $ImportTestToolkitToBcContainer,
     [scriptblock] $CompileAppInBcContainer,
+    [Parameter(ParameterSetName='UseCompilerFolder')]
+    [scriptblock] $PreCompileApp,
+    [Parameter(ParameterSetName='UseCompilerFolder')]
+    [scriptblock] $PostCompileApp,
     [scriptblock] $GetBcContainerAppInfo,
     [scriptblock] $PublishBcContainerApp,
     [scriptblock] $UnPublishBcContainerApp,
@@ -710,6 +721,15 @@ if ($CompileAppInBcContainer) {
 else {
     $CompileAppInBcContainer = { Param([Hashtable]$parameters) Compile-AppInBcContainer @parameters }
 }
+
+if ($PreCompileApp) {
+    Write-Host -ForegroundColor Yellow "Custom pre-compilation script defined."; Write-Host $PreCompileApp.ToString()
+}
+
+if($PostCompileApp) {
+    Write-Host -ForegroundColor Yellow "Custom post-compilation script defined."; Write-Host $PostCompileApp.ToString()
+}
+
 if ($GetBcContainerAppInfo) {
     Write-Host -ForegroundColor Yellow "GetBcContainerAppInfo override"; Write-Host $GetBcContainerAppInfo.ToString()
 }
@@ -1746,11 +1766,34 @@ Write-Host -ForegroundColor Yellow @'
         }
     }
 
+    # Script block for compiling an app using a compiler folder
+    $CompileAppWithBCCompilerFolder = {
+        Param(
+            [scriptblock] $PreCompileApp,
+            [scriptblock] $PostCompileApp,
+            $Parameters
+        )
+
+        if($PreCompileApp) {
+            Write-Host "Running custom PreCompileApp"
+            Invoke-Command -ScriptBlock $PreCompileApp -ArgumentList $Parameters
+        }
+
+        $appFile = Compile-AppWithBcCompilerFolder @Parameters
+
+        if($PostCompileApp) {
+            Write-Host "Running custom PostCompileApp"
+            Invoke-Command -ScriptBlock $PostCompileApp -ArgumentList $Parameters
+        }
+        
+        return $appFile
+    }
+
     try {
         Write-Host "`nCompiling $($Parameters.appProjectFolder)"
         $withCops = $Parameters+$CopParameters
         if ($useCompilerFolder) {
-            $appFile = Compile-AppWithBcCompilerFolder @parameters
+            $appFile = Invoke-Command -ScriptBlock $CompileAppWithBCCompilerFolder -ArgumentList $PreCompileApp, $PostCompileApp, $Parameters
         }
         else {
             $appFile = Invoke-Command -ScriptBlock $CompileAppInBcContainer -ArgumentList $withCops
@@ -1760,7 +1803,7 @@ Write-Host -ForegroundColor Yellow @'
         if ($escapeFromCops) {
             Write-Host "Retrying without Cops"
             if ($useCompilerFolder) {
-                $appFile = Compile-AppWithBcCompilerFolder @parameters
+                $appFile = Invoke-Command -ScriptBlock $CompileAppWithBCCompilerFolder -ArgumentList $PreCompileApp, $PostCompileApp, $Parameters
             }
             else {
                 $appFile = Invoke-Command -ScriptBlock $CompileAppInBcContainer -ArgumentList $Parameters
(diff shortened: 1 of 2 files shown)


AppHandling/Run-AlPipeline.ps1
```diff
@@ -710,6 +721,15 @@ if ($CompileAppInBcContainer) {
 else {
     $CompileAppInBcContainer = { Param([Hashtable]$parameters) Compile-AppInBcContainer @parameters }
 }
+
+if ($PreCompileApp) {
+    Write-Host -ForegroundColor Yellow "Custom pre-compilation script defined."; Write-Host $PreCompileApp.ToString()
+}
+
+if($PostCompileApp) {
```
mazhelez, line 729:
```suggestion
if ($PostCompileApp) {

```

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,11 +1766,34 @@ Write-Host -ForegroundColor Yellow @'
...
+    # Script block for compiling an app using a compiler folder
+    $CompileAppWithBCCompilerFolder = {
+        Param(
+            [scriptblock] $PreCompileApp,
+            [scriptblock] $PostCompileApp,
+            $Parameters
+        )
+
+        if($PreCompileApp) {
+            Write-Host "Running custom PreCompileApp"
```
mazhelez, line 1778:
```suggestion
            Write-Host "Running custom pre-compilation script..."

```

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,11 +1766,34 @@ Write-Host -ForegroundColor Yellow @'
...
+            Invoke-Command -ScriptBlock $PreCompileApp -ArgumentList $Parameters
+        }
+
+        $appFile = Compile-AppWithBcCompilerFolder @Parameters
+
+        if($PostCompileApp) {
+            Write-Host "Running custom PostCompileApp"
```
mazhelez, line 1785:
```suggestion
            Write-Host "Running custom post-compilation script..."

```

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,11 +1766,34 @@ Write-Host -ForegroundColor Yellow @'
...
+        }
+        
+        return $appFile
+    }
+
     try {
         Write-Host "`nCompiling $($Parameters.appProjectFolder)"
         $withCops = $Parameters+$CopParameters
         if ($useCompilerFolder) {
-            $appFile = Compile-AppWithBcCompilerFolder @parameters
```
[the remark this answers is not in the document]
freddydk: Yes - it should

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,11 +1767,36 @@ Write-Host -ForegroundColor Yellow @'
...
+        Param(
+            [scriptblock] $PreCompileApp,
+            [scriptblock] $PostCompileApp,
+            $Parameters
+        )
+
+        if($PreCompileApp) {
+            Write-Host "Running custom pre-compilation script..."
+
+            Invoke-Command -ScriptBlock $PreCompileApp -ArgumentList $Parameters
```
freddydk, line 1781: I don't think we should transfer all parameters to precompile - only the ones we need
e..g the PostCompile should probably also have the appfile as a parameter.

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,11 +1767,36 @@ Write-Host -ForegroundColor Yellow @'
...
+        
+        return $appFile
+    }
+
     try {
         Write-Host "`nCompiling $($Parameters.appProjectFolder)"
         $withCops = $Parameters+$CopParameters
         if ($useCompilerFolder) {
-            $appFile = Compile-AppWithBcCompilerFolder @parameters
+            $appFile = Invoke-Command -ScriptBlock $CompileAppWithBCCompilerFolder -ArgumentList $PreCompileApp, $PostCompileApp, $Parameters
```
freddydk, line 1799: We need to call the same pre and post when doing Compile-AppInBcContainer

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1760,7 +1806,7 @@ Write-Host -ForegroundColor Yellow @'
         if ($escapeFromCops) {
             Write-Host "Retrying without Cops"
             if ($useCompilerFolder) {
-                $appFile = Compile-AppWithBcCompilerFolder @parameters
+                $appFile = Invoke-Command -ScriptBlock $CompileAppWithBCCompilerFolder -ArgumentList $PreCompileApp, $PostCompileApp, $Parameters
```
freddydk, line 1809: Not sure this is the right approach - this means that you might get pre and post called twice for the same app.
I think you should call pre before compiling - then compile and maybe without cops and then call post.

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
mazhelez: Why so? For `Compile-AppInBcContainer` there is already `CompileAppInBcContainer` override. It cannot be the same script (it should differ based on whether `useCompilerFolder` is on).

`PreCompileApp` and `PostCompileApp` are only expected when `useCompilerFolder` is on.

AppHandling/Run-AlPipeline.ps1
mazhelez: `CompileAppInBcContainer` takes all parameters as input. What are your concerns?

AppHandling/Run-AlPipeline.ps1
mazhelez: Ok, my understanding was that the pre and post compile scripts are to cover the scenarios when `useCompilerFolder` is on and the custom `CompileAppInBcContainer` won't be called.

In any case, the only way one can use the same scripts for both when `useCompilerFolder` is on and off is to keep checking the value of the flag.
e.g.
```
if ($useCompilerFolder`) {
    # do something in the compiler folder
} 
else {
    # do something in the BC container
}
```

AppHandling/Run-AlPipeline.ps1
mazhelez: Now pre and post compile scripts are called no matter if `useCompilerFolder` is on or off.

For now, we pass all the compilation parameters to both scripts (even as a _ref_ to the pre compile script) + the .app file path to the post compile script.

AppHandling/Run-AlPipeline.ps1
mazhelez: Now the scripts are called only once.

AppHandling/Run-AlPipeline.ps1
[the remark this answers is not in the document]
mazhelez: Done.

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1635,6 +1654,9 @@ Write-Host -ForegroundColor Yellow @'
         "appProjectFolder" = $folder
         "appOutputFolder" = $appOutputFolder
         "appSymbolsFolder" = $appPackagesFolder
+        "isApp" = $app
+        "isTestApp" = $testApp
```
freddydk, line 1658: Is that a parameter on Compile-AppInBcContainer ?
Won't this fail when calling that?
Maybe create another hashtable for event specific parameters like this.
and then add the appFile after compilation - I assume in the post compile step you want the appFile right?

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,31 +1768,47 @@ Write-Host -ForegroundColor Yellow @'
...
         else {
             throw $_
         }
     }
 
+    # Run post-compile script if specified
+    if($PostCompileApp) {
+        Write-Host "Running custom post-compilation script..."
+
+        Invoke-Command -ScriptBlock $PostCompileApp -ArgumentList $appFile, $compilationParams
```
freddydk, line 1809: What does the param block in the post and precompile app look like?
Is that a HashTable like the other overrides ?

AppHandling/Run-AlPipeline.ps1
mazhelez: `$app` is otherwise a global variable for (see [this reference](https://github.com/microsoft/ALAppExtensions/blob/25e371e96f2651dabd2af08df11f8d3d5465d510/Build/Scripts/CompileAppInBcContainer.ps1#LL11C1-L12C9) for example).

My idea of adding the values of `$app`, `$testApp` and `$bcptTestApp` as parameters was to avoid using variables that were defined in `Run-AlPipeline.ps1` in the override scripts.

But you are right, it is breaking to simply add extra parameters.
I will give it an extra thought.

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,31 +1784,54 @@ Write-Host -ForegroundColor Yellow @'
...
         }
         else {
-            $appFile = Invoke-Command -ScriptBlock $CompileAppInBcContainer -ArgumentList $withCops
+            $appFile = Invoke-Command -ScriptBlock $CompileAppInBcContainer -ArgumentList $compilationParams
         }
     }
     catch {
         if ($escapeFromCops) {
             Write-Host "Retrying without Cops"
+            $compilationParams = $Parameters
```
freddydk, line 1815: Won't this override any changes made to compilationParams by the pre-event?
Maybe CopParameters should be added in like 1809 instead amd compilationParams kept the same as modified by the pre-event?

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
```diff
@@ -174,6 +176,31 @@
...
   Override function parameter for Compile-AppInBcContainer
+ .Parameter PreCompileApp
+  Custom script to run before compiling an app.
+  The script should accept the type of the app and a reference to the compilation parameters.
+  Possible values for $appType are: app, testApp, bcptApp
+  Example:
+  {
+    param(
+        [string] $appType,
+        [ref] $compilationParams
```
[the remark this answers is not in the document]
freddydk: No, it shouldn't

AppHandling/Run-AlPipeline.ps1
```diff
@@ -174,6 +176,31 @@
...
+    )
+    ...
```
freddydk, line 189: Add an example on how to use compilationParams here:

$path = $compilationParams.Value.appProjectFolder

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,31 +1784,58 @@ Write-Host -ForegroundColor Yellow @'
...
         if ($escapeFromCops) {
             Write-Host "Retrying without Cops"
+
+            # Remove Cops parameters
+            $compilationParamsCopy = $compilationParams.Clone()
+            $compilationParamsCopy.Keys | Where-Object {$_ -in $CopParameters.Keys} | ForEach-Object { $compilationParams.Remove($_) }
+
             if ($useCompilerFolder) {
-                $appFile = Compile-AppWithBcCompilerFolder @parameters
+                $appFile = Compile-AppWithBcCompilerFolder @compilationParams
```
freddydk, line 1821: use the Copy?

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
freddydk: No

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1746,31 +1784,58 @@ Write-Host -ForegroundColor Yellow @'
...
         else {
             throw $_
         }
     }
 
+    # Run post-compile script if specified
+    if($PostCompileApp) {
+        Write-Host "Running custom post-compilation script..."
+
+        Invoke-Command -ScriptBlock $PostCompileApp -ArgumentList $appFile, $appType, $compilationParams
```
freddydk, line 1836: Now this one contains a different parameter set whether or not the cops where removed
Is this on purpose?

AppHandling/Run-AlPipeline.ps1
mazhelez: Isn't that more like how to use `ref`? As in [about_Ref](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_ref?view=powershell-5.1).

I thought of adding some doc around what parameters are available (e.g. `appProjectFolder`, `appOutputFolder` etc.) these were never documented anyway.

AppHandling/Run-AlPipeline.ps1
mazhelez: Well, yeah. 

The idea was the post compile script to get the _actual_ compilation parameters that were used to create the app file.

AppHandling/Run-AlPipeline.ps1
mazhelez: Added an example in the parameter's doc.

PR #3117, opened 2023-07-12
add logic for using update window at app installation

JeremiasBr:
Hi,

i added the option to use the switch "**useEnvironmentUpdateWindow**" in "**Saas/Install-BcAppFromAppSource.ps1**" and changed the scripts logic accordingly.
Using this switch only schedules the install operation and - of course - does not wait for the operation to be finished.

Also added comment in release notes.

Greetings,
Jeremias


freddydk:
Thanks, looks good

PR #3169, opened 2023-08-30
feat: Adds new parameters to Start-BcContainerAppDataUpgrade

fridrichovsky:
Provide same parameters from Start-NAVAppDataUpgrade to Start-BcContainerAppDataUpgrade.

### DIFF
--- a/AppHandling/Start-NavContainerAppDataUpgrade.ps1
+++ b/AppHandling/Start-NavContainerAppDataUpgrade.ps1
@@ -23,7 +23,7 @@ function  Start-BcContainerAppDataUpgrade {
         [string] $appVersion = "",
         [string] $language = "",
         [string] $exclusiveAccessTicket = "",
-        [string] $path = "",
+        [string] $path,
         [switch] $force,
         [switch] $skipVersionCheck,
         [ValidateSet("Add", "Clean", "Development", "ForceSync", "None")]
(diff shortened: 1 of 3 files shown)


AppHandling/Start-NavContainerAppDataUpgrade.ps1
```diff
@@ -18,39 +18,59 @@ function  Start-BcContainerAppDataUpgrade {
...
+                $parameters += @{ "SyncMode" = $syncMode }
+            }
+            if ($force.IsPresent) {
+                $parameters += @{ "Force" = $force }
+            }
+            if ($skipVersionCheck.IsPresent) {
+                $parameters += @{ "SkipVersionCheck" = $skipVersionCheck }
+            }
+            Start-NAVAppDataUpgrade @parameters
+        } -ArgumentList $appName, $appVersion, $tenant, $language, $exclusiveAccessTicket, $path, $syncMode, $force, $skipVersionCheck
```
freddydk, line 64: path?

AppHandling/Start-NavContainerAppDataUpgrade.ps1
```diff
@@ -18,39 +18,59 @@ function  Start-BcContainerAppDataUpgrade {
     Param (
         [string] $containerName = $bcContainerHelperConfig.defaultContainerName,
         [string] $tenant = "default",
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [string] $appName,
         [string] $appVersion = "",
-        [string] $language = ""
+        [string] $language = "",
```
freddydk, line 24: path?
and what is path? if that should be a path specification, should it be a specification inside the container or outside?
Use Get-BcContainerPath -containerName -path to get the shared path of a path on the host if this is what's needed.

freddydk (changes requested):

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: I was surprised that I can set path, but yes. This is from documentation.
Start-NAVAppDataUpgrade
     [-ServerInstance] <String>
     [-Tenant <TenantId>]
     -Path <String>
     [-Language <LanguageSetting>]
     [-ExclusiveAccessTicket <String>]
     [-PassThru]
     [-SkipVersionCheck]
     [-SyncMode <NavAppSyncMode>]
     [-Force]
     [<CommonParameters>]

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: You are right I forgot that I am in conatiner

fridrichovsky: I tried change path like Publish-BCContainerApp.ps1, but I am not sure if it is ok this way.

freddydk: Most places where we have this - we just use

```
Invoke-ScriptInBcContainer -containername $containername -scriptblock { Param($containerPath)

code

} -argumentList (Get-BcContainerPath -containerName $containerName -path $path -throw)
```

Then people will need to place the files in a location, which is shared with the container - but that can easily be done by a simple copy to the container - or share the folder with the container.

I think we should do this

freddydk (changes requested):

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: Please check if it is OK.

AppHandling/Start-NavContainerAppDataUpgrade.ps1
```diff
@@ -18,39 +18,64 @@ function  Start-BcContainerAppDataUpgrade {
...
         [string] $containerName = $bcContainerHelperConfig.defaultContainerName,
         [string] $tenant = "default",
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [string] $appName,
         [string] $appVersion = "",
-        [string] $language = ""
+        [string] $language = "",
+        [string] $exclusiveAccessTicket = "",
+        [string] $path,
```
freddydk, line 26: I prefer defaulting parameters which are not mandatory.
Set path and syncmode to ""

AppHandling/Start-NavContainerAppDataUpgrade.ps1
```diff
@@ -18,39 +18,64 @@ function  Start-BcContainerAppDataUpgrade {
...
-        $parameters = @{
-            "ServerInstance" = $ServerInstance;
-            "Name" = $appName;
-            "Tenant" = $tenant
-        }
-        if ($appVersion) {
-            $parameters += @{ "Version" = $appVersion }
+    $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
+    try {
+        $containerPath = $null
```
freddydk, line 35: use empty string instead of null - inside the container it will be empty (not null)

freddydk (changes requested):
A few fixes more...

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: OK

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: OK

AppHandling/Start-NavContainerAppDataUpgrade.ps1
```diff
@@ -18,39 +18,64 @@ function  Start-BcContainerAppDataUpgrade {
...
+                $parameters += @{ "Path" = $containerPath }
+            }
+            if ($syncMode) {
+                $parameters += @{ "SyncMode" = $syncMode }
+            }
+            if ($force.IsPresent) {
+                $parameters += @{ "Force" = $force }
+            }
+            if ($skipVersionCheck.IsPresent) {
+                $parameters += @{ "SkipVersionCheck" = $skipVersionCheck }
```
freddydk, line 66: I think these needs to be 
"skipVersionCheck" = $true
(same with force and exclusiveAccessTicket)

freddydk (changes requested):

AppHandling/Start-NavContainerAppDataUpgrade.ps1
fridrichovsky: -exclusiveAccessTicket is string value 
-force and -skipVersionCheck are in default set to $false. It means that your function works now with these defaults. Do you want change defaults? It going to change function working for all who use this function without new paramaters.

freddydk: The only change I want is this:

```
if ($force.IsPresent) {
    $parameters += @{ "Force" = $true }
}
if ($skipVersionCheck.IsPresent) {
    $parameters += @{ "SkipVersionCheck" = $true}
}
```

mostly because force is of type System.Management.Automation.SwitchParameter and I am always in doubt whether the other construct works...
It does - I tried, but to improve readability...

fridrichovsky: I catch the point now :-). Sorry for my misunderstanding. I changed it.

freddydk (approved):

freddydk:
Thanks

PR #3177, opened 2023-09-04
add Switch "getCompanyInfo" to "New-BcEnvironment"

JeremiasBr:
Hi,

I added the switch "getCompanyInfo" to the function "New-BcEnvironment".
This allows users to prevent the script from getting environment information like companies via the Automation API after the environment was created.

Without the switch, the Automation API is called right after the environment was created, which results in an error, since the application used for authentication via S2S has not been set in Business Central yet.
This enables using S2S - Authentication for environment creation.

Greetings,
Jeremias


freddydk:
As I read the code, this switch makes New-BcEnvironment change functionality - right?
Please change the switch to -doNotGetCompanyInfo and reverse the checks
Else everybody using impersonation will have to change their code.

JeremiasBr:
Thank you - I changed the script accordingly.

freddydk (approved):

PR #3179, opened 2023-09-06
Fixing typo in the code

kine:
There is typo in the code, leading to this error when running Start-BcContainerAppDataUpgrade:

```
The term '=' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

BCContainerhelper version> 5.0.6-preview1069


freddydk:
Thanks Kamil

freddydk (approved):

PR #3183, opened 2023-09-08
Added parameter copyCustomCodeCopsToAnalyzersFolder

MODUSCarstenScholling:
Added the parameter copyCustomCodeCopsToAnalyzersFolder to allow to copy custom analyzer dlls to the containers analyzers folder. This makes sure all referenced dlls are in the same place and can be found by custom analyzers.

Parameter CustomCodeCops has been expanded to allow patterns and can also pass non-dll files (configuration). But only dll files are added as custom cops.

Handling of custom cops has been moved below extracting vsix with alc.

Example

```PowerShell
# [...]
$customCops = @("C:\ProgramData\BcContainerHelper\Extensions\bc230de\compile\MC.Dynamics.Bc.Al.ModusCop.*");

Compile-AppInBcContainer -containerName bc230de -tenant default -appProjectFolder $appProjectFolder -appOutputFolder $outputFolder -appSymbolsFolder  $appSymbolsFolder `
    -basePath $sharedBasePath -AzureDevOps -CustomCodeCops $customCops -copyCustomCodeCopsToAnalyzersFolder -credential $bcCredential;
```

Sample Output

```
Using Symbols Folder: C:\ProgramData\BcContainerHelper\Extensions\bc230de\compile\OurApp\.alpackages
Copied custom code cop MC.Dynamics.Bc.Al.ModusCop.dll to analyzers folder.
Processing dependency Microsoft_Application_20.1.0.0 ()
Dependency App exists
Processing dependency Microsoft_System_20.0.0.0 ()
Dependency App exists
Processing dependency MODUS Consult GmbH_BaseStuff_20.4.23081401.0 (xxx)
Dependency App exists
Compiling...
.\alc.exe /project:"C:\ProgramData\BcContainerHelper\Extensions\bc230de\compile\OurApp\app" /packagecachepath:"C:\ProgramData\BcContainerHelper\Extensions\bc230de\compile\OurApp\.alpackages" /out:"C:\ProgramData\
BcContainerHelper\Extensions\bc230de\compile\out\OurApp_20.1.23050900.0 .app" /analyzer:C:\build\vsix\extension\bin\Analyzers\MC.Dynamics.Bc.Al.ModusCop.dll /BuildBy:"BcContainerHelper,5.0.5" /assem
blyprobingpaths:"C:\Program Files\dotnet\shared","C:\ProgramData\BcContainerHelper\Extensions\bc230de\.netPackages\Service"
```

### DIFF
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -46,6 +46,10 @@
   Add this switch to Enable External Rulesets
  .Parameter CustomCodeCops
   Add custom AL code Cops when compiling apps.
+  You can specify a pattern like "Cop.*" to allow additional files not ending with *.dll. This is helpful if used with copyCustomCodeCopsToAnalyzersFolder.
+ .Parameter copyCustomCodeCopsToAnalyzersFolder
+  Add this switch to copy the passed CustomCodeCops into the alc analyzers folder inside the container.
+  All files passed in CustomCodeCops will be copied, but only *.dll files are added as an analyzer.
  .Parameter Failon
   Specify if you want Compilation to fail on Error or Warning
  .Parameter nowarn
@@ -113,6 +117,7 @@ function Compile-AppInBcContainer {
         [string] $rulesetFile,
         [switch] $enableExternalRulesets,
         [string[]] $CustomCodeCops = @(),
+        [switch] $copyCustomCodeCopsToAnalyzersFolder,
         [Parameter(Mandatory=$false)]
         [string] $nowarn,
         [string[]] $preProcessorSymbols = @(),
@@ -216,18 +221,6 @@ try {
             throw "The rulesetFile ($rulesetFile) is not shared with the container."
         }
     }
-    
-    $CustomCodeCopFiles = @()
-    if ($CustomCodeCops.Count -gt 0) {
-        $CustomCodeCops | ForEach-Object {
-            $customCopPath = Get-BcContainerPath -containerName $containerName -path $_
-            if ("$customCopPath" -eq "") {
-                throw "The custom code cop ($_) is not shared with the container."
-            }
-
-            $CustomCodeCopFiles += $customCopPath
-        }
-    }
 
     if (!(Test-Path $appOutputFolder -PathType Container)) {
         New-Item $appOutputFolder -ItemType Directory | Out-Null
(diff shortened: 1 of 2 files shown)


freddydk:
We would need the same functionality implemented in Compile-AppWithBcCompilerFolder (which over time will be the primary function used for compiling apps)

freddydk:
+ a line in ReleaseNotes explaining this.

MODUSCarstenScholling:
Release note added.

MODUSCarstenScholling:
> We would need the same functionality implemented in Compile-AppWithBcCompilerFolder (which over time will be the primary function used for compiling apps)

Hmm, thinking of New-BcCompilerFolder for implementation. But the folder structure is fixed, so it is possible to use a simple Copy-Item then. I tend to add this to New-BcCompilerFolder.

Would allowing a placeholder like ${AnalyzerFolder} for Compile-AppWithBcCompilerFolder and CustomCodeCops make sense? Or a useWorkspaceCodeAnalyzers to re-use existing settings?

Our settings.json looks like this:

```json
{
	"al.enableCodeAnalysis": true,
	"al.codeAnalyzers": [
		"${CodeCop}",
		"${UICop}",
		"${AppSourceCop}",
		"${AnalyzerFolder}\\MC.Dynamics.Bc.Al.ModusCop.dll"
	]
}
```

freddydk:
> Hmm, thinking of New-BcCompilerFolder for implementation. But the folder structure is fixed, so it is possible to use a simple Copy-Item then. I tend to add this to New-BcCompilerFolder.

Primary reason for the folder structure in New-BcCompilerFolder is, that it is used for caching in AL-Go for GitHub. This means that there shouldn't be any instance specific stuff added to the cache folder.

MODUSCarstenScholling:
> Primary reason for the folder structure in New-BcCompilerFolder is, that it is used for caching in AL-Go for GitHub. This means that there shouldn't be any instance specific stuff added to the cache folder.

I'm not yet in that AL-Go stuff, but if caching means having something fully setup and everything in place sitting and waiting for work, additional cops are really something that would belong to this cache.

For Microsoft this is Code/AppSource/UICop, for us and others it is eventually MODUSCop, LinterCop, any other cop in addition raising diagnostic for an app. So, this would not be instance specific but company or account specific.

> Not sure what you mean with this. The CompilerFolder feature doesn't really have anything to do with VS Code (i.e. settings.json) - it is just container-less compiling really (to save time).

Understood. What I meant was a feature allowing to pass an array of cops but allowing to reference "${AnalyzerFolder}" which is to be replaced by the actual folder when compiling. Only an idea.

Would it be possible to accept the PR without the changes for AL-Go? I promise an additional PR for it, based on our discussion outcome. But that may take 2-3 weeks.

MODUSCarstenScholling:
I am still on this, but regular work is keeping me busy until the new year. I will come back to this after vacation.

PR #3186, opened 2023-09-14
Add generateErrorLog setting to Run-ALPipeline

mazhelez:
Add generateErrorLog setting to `Run-ALPipeline`

The setting is propagated down to `Compile-AppInNavContainer` and `Compile-AppWithBcCompilerFolder` and is used to set the compiler setting `errorLog`.

The generated error log file has the same name as the `.app` file but the file extension is `.errorLog.json`.


AppHandling/Run-AlPipeline.ps1
```diff
@@ -13,6 +13,8 @@
   License file to use for AL Pipeline.
  .Parameter containerName
   This is the containerName going to be used for the build/test container. If not specified, the container name will be the pipeline name followed by -bld.
+ .Parameter generateErrorLog
+  Path to file where to output the alerts from the compiler.
```
aholstrup1, line 17:
```suggestion
  Include this switch if compiler should output the errorlog as a file
```

AppHandling/Run-AlPipeline.ps1
```diff
@@ -2426,6 +2431,28 @@ if (!(Test-Path $destFolder -PathType Container)) {
...
+# Copy error logs to build artifact folder
+if($generateErrorLog) {
+    # Define destination folder for error logs
+    $destFolder = Join-Path $buildArtifactFolder "ErrorLogs"
+    if (!(Test-Path $destFolder -PathType Container)) {
+        New-Item $destFolder -ItemType Directory | Out-Null
+    }
+
+    # Copy error logs for all apps
+    @($apps, $testApps, $bcptTestApps) | ForEach-Object {
```
aholstrup1, line 2444: Is it only apps that successfully compiled that are in $apps, $testapps and $bcptTestApps? If an app fails to compile shouldn't we still get the errorlog?

mazhelez: That's a good point, but I'm not sure the log file is generated if the app fails compiling. I'll check.

CompilerFolderHandling/Compile-AppWithBcCompilerFolder.ps1
```diff
@@ -382,6 +383,11 @@ try {
         $alcParameters += @("/nowarn:$nowarn")
     }
 
+    if ($generateErrorLog) {
+        $errorLogFilePath = $appOutputFile -replace ".app$", ".errorLog.json"
```
freddydk, line 387: Single quotes? (when a $ sign is present)

AppHandling/Run-AlPipeline.ps1
```diff
@@ -13,6 +13,8 @@
   License file to use for AL Pipeline.
  .Parameter containerName
   This is the containerName going to be used for the build/test container. If not specified, the container name will be the pipeline name followed by -bld.
+ .Parameter generateErrorLog
+  Switch parameter on whether to generate an alerts log file. Default is false.
```
freddydk, line 19: Add to the comment that an alerts log file is generated for each compiled app file and placed in the output folder next to the app file.
(I think that is what it does right?)

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
mazhelez: Done, now the file are copied always.

freddydk (approved):

PR #3200, opened 2023-09-27
Add new parameter to Publish-PerTenantExtensionApps

vpedraja:
This implementation aims to include a new parameter on the Publish-PerTenantExtensionApps script in order to avoid the extension list output to show up in the psh console whenever a user runs the function. The default behavior would be to show that extension list before and after the extension installation but if the user wants to use this new parameter, they should be able to and the Write-Host with this data would not execute.


vpedraja:

freddydk (approved):

PR #3213, opened 2023-10-20
Latest insider artifacts fail building in compilerfolder

aholstrup1:
Latest insider artifacts fail building in compilerfolder

`AL-Go action ran: RunPipeline Telemetry Correlation Id: 5f39d8a8-e12d-4443-bbe1-9ffe5b053643
Error: Unexpected error when running action. Error Message: Exception calling "Create" with "2" argument(s): "Could not load file or assembly 'System.IO.Packaging, Version=4.0.5.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.", StackTrace: at GetAppInfo, C:\ProgramData\BcContainerHelper\6.0.1-preview1082\BcContainerHelper\HelperFunctions.ps1: line 1154 <- at Compile-AppWithBcCompilerFolder, C:\ProgramData\BcContainerHelper\6.0.1-preview1082\BcContainerHelper\CompilerFolderHandling\Compile-AppWithBcCompilerFolder.ps1: line 178 `

### DIFF
--- a/AppHandling/Compile-AppInNavContainer.ps1
+++ b/AppHandling/Compile-AppInNavContainer.ps1
@@ -500,6 +500,7 @@ try {
                                 $alcPath = 'C:\build\vsix\extension\bin'
                                 Add-Type -Path (Join-Path $alcPath Newtonsoft.Json.dll)
                                 Add-Type -Path (Join-Path $alcPath System.Collections.Immutable.dll)
+                                Add-Type -Path (Join-Path $alcPath System.IO.Packaging.dll)
                                 Add-Type -Path (Join-Path $alcPath Microsoft.Dynamics.Nav.CodeAnalysis.dll)
 
                                 $packageStream = [System.IO.File]::OpenRead($symbolsFile)
--- a/HelperFunctions.ps1
+++ b/HelperFunctions.ps1
@@ -1147,6 +1147,7 @@ function GetAppInfo {
                     Add-Type -AssemblyName System.Text.Encoding
                     LoadDLL -Path (Join-Path $alcDllPath Newtonsoft.Json.dll)
                     LoadDLL -Path (Join-Path $alcDllPath System.Collections.Immutable.dll)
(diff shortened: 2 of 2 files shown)


freddydk:
Will this work with all old versions of BC artifacts as well?

freddydk:
Also, there is a similar pattern in Compile-AppInBcContainer line 502 - should the DLL be loaded there as well?

aholstrup1:
> Will this work with all old versions of BC artifacts as well? I mean is that DLL available in the vsix on BC 20.0 artifacts f.ex.?

I don't know if it's available in **all** old versions of BCArtifacts. I checked 20.5.45456.61518. That one has the dll so it seems like the DLL has been there for at least some years

aholstrup1:
> Also, there is a similar pattern in Compile-AppInBcContainer line 502 - should the DLL be loaded there as well?
I'll check what happens if I run the build with a container (i.e. not with compilerfolder). I guess you're right. I'll probably have to add it in there as well

freddydk:
> > Will this work with all old versions of BC artifacts as well? I mean is that DLL available in the vsix on BC 20.0 artifacts f.ex.?
> 
> I don't know if it's available in **all** old versions of BCArtifacts. I checked 20.5.45456.61518. That one has the dll so it seems like the DLL has been there for at least some years

I checked version 16 - it is there.
I assume that we will have to add it in both places and then we should be good

aholstrup1:
> > > Will this work with all old versions of BC artifacts as well? I mean is that DLL available in the vsix on BC 20.0 artifacts f.ex.?
> > 
> > 
> > I don't know if it's available in **all** old versions of BCArtifacts. I checked 20.5.45456.61518. That one has the dll so it seems like the DLL has been there for at least some years
> 
> I checked version 16 - it is there. I assume that we will have to add it in both places and then we should be good

After testing it seems that it works both with and without the change to Compile-AppInNavContainer.ps1. Should I remove it again? Or you want to merge it as-is?

freddydk (approved):

PR #3239, opened 2023-11-21
adding sqlTimeout par to Restore-BcDatabaseFromArtifacts

lktraser:
This parameter is useful to restore large databases to external SQL without it the command will just timeout.

I don't know if the default value makes sense here I just copied it over. In my case I just want to max it out.


Bacpac/Restore-BcDatabaseFromArtifacts.ps1
```diff
@@ -140,7 +143,7 @@ try {
             }
 
             Write-Host "Restoring database $dbName into $destinationPath"
-            New-NAVDatabase -DatabaseServer $databaseServer -DatabaseInstance $databaseInstance -DatabaseName $dbName -FilePath $bakFile -DestinationPath $destinationPath | Out-Null
+            New-NAVDatabase -DatabaseServer $databaseServer -DatabaseInstance $databaseInstance -DatabaseName $dbName -FilePath $bakFile -DestinationPath $destinationPath -timeout $sqlTimeout | Out-Null
```
freddydk, line 146: This code is running in a background job (Start-Job in line 60) and argumentlist in line 206.
You have to transfer the sqlTimeout value to that job - else it is not defined.

freddydk (changes requested):

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
lktraser: Absolutely, I had it hardcoded and forgot to add it to the arguments. Updated the argument lists

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
```diff
@@ -36,7 +38,8 @@ function Restore-BcDatabaseFromArtifacts {
         [Parameter(Mandatory=$false)]
         [string] $bakFile,
         [switch] $multitenant,
-        [switch] $async
+        [switch] $async,
+        [int] $sqlTimeout = 300
```
freddydk, line 42: I cannot find the right value of the default timeout.
The documentation says "none", meaning that it might be up to the server, so we should probably avoid adding the parameter if not specified.
Like setting the default to -1 here and then modify the code for calling New-NavDatabase to:

```
$params = @{
    "DatabaseServer" = $databaseServer
    "DatabaseInstance" = $databaseInstance
    "DatabaseName" = $dbName
    "FilePath" = $bakFile
    "DestinationPath" = $destinationPath
}
if ($sqlTimeout -ge 0) {
    $params += @{ "timeout" = $sqlTimeout }
}
New-NavDatabase @params | Out-Null
```

That would ensure that the timeout isn't set if not provided.

Make sense?

freddydk (changes requested):

lktraser:
@microsoft-github-policy-service agree company="TRASER Software GmbH"

freddydk (approved):

freddydk:
Thanks Leon

Issue #3235, opened 2023-11-17
UseCompilerFolder with Azure DevOps Pipeline - Warnings and Errors Shown as Plain Text


MattTraxinger:
When using Run-AlPipeline with a compiler folder in an Azure DevOps pipeline, warnings appear to be output as plain text. Errors are the same, but since the result contains "App Generation Failed" the pipeline at least still fails. FailOn Warning also does not appear to work in this scenario.

I believe the result returned from the compiler is being treated as one large block of text and failing to meet any of the regex conditions in Convert-ALCOutputToAzureDevOps, as opposed to being processed line by line, but I could be mistaken. This is just what I observed when copying the code out and running it myself.

I've taken out the irrelevant parts of the output for readability, but am happy to provide the whole thing if needed.

```
2023-11-17T17:32:35.9916164Z ##[section]Starting: Run Pipeline
2023-11-17T17:32:36.0049977Z ==============================================================================
2023-11-17T17:32:36.0050316Z Task         : PowerShell
2023-11-17T17:32:36.0050469Z Description  : Run a PowerShell script on Linux, macOS, or Windows
2023-11-17T17:32:36.0050803Z Version      : 2.231.0
2023-11-17T17:32:36.0050947Z Author       : Microsoft Corporation
2023-11-17T17:32:36.0051163Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/powershell
2023-11-17T17:32:36.0051597Z ==============================================================================
2023-11-17T17:32:37.2463062Z Generating script.
2023-11-17T17:32:37.3092459Z Formatted command: . '[AgentDriveMapping]\426\s\scripts\DevOps\DevOps-Pipeline.ps1' -environment "AzureDevOps" -appRevision 106971
2023-11-17T17:32:37.3458014Z ========================== Starting Command Output ===========================
2023-11-17T17:32:37.3666616Z ##[command]"C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". '[AgentDriveMapping]\_temp\0052bad3-a7cb-485a-af34-bc2a0bad3828.ps1'"
2023-11-17T17:32:39.0414616Z 
2023-11-17T17:32:39.3776846Z BcContainerHelper version 6.0.0
2023-11-17T17:32:39.8300537Z Setting defaultNewContainerParameters = @{Isolation=hyperv}
[...]

PR #3250, opened 2023-11-27
Fix CompileResult Format in Compile-AppWithBcCompilerFolder

PeterDurrer:
Remove Out-String for alcCmd return result for correct working Convert-ALCOutputToAzureDevOps
@freddydk this is PullRequest for #3235


freddydk:
Thanks - not sure why that out-string was there in the first place.
Everything in that function assumes that $result is an array (which Out-String destroys).

Thanks

freddydk (approved):

PR #3264, opened 2023-12-05
NuGet changes

freddydk:
Rename appinfo cache from AppInfo.json to cache_AppInfo.json (in order for it to be picked up by .gitignore)
Proof of concept functionality for NuGet feeds:
- Add trusted feeds to ContainerHelper settings.
- Search all trusted feeds for packages in Get-BcNuGetPackage
- Some refactoring of NuGet code
- Allow searching for Earliest, Latest, Any or Exact version of NuGet package
- Support NuGet Versioning schema


NuGet/New-BcNuGetPackage.ps1
```diff
@@ -129,51 +118,45 @@ Function New-BcNuGetPackage {
...
             $XmlObjectWriter.WriteEndElement()
         }
         $XmlObjectWriter.WriteEndElement()
         $XmlObjectWriter.WriteEndElement()
+        $XmlObjectWriter.WriteStartElement("files")
+        $XmlObjectWriter.WriteStartElement("file")
+        $appFileName = [System.IO.Path]::GetFileName($appfile)
+        $XmlObjectWriter.WriteAttributeString("src", $appFileName );
+        $XmlObjectWriter.WriteAttributeString("target", $appFileName);
+        $XmlObjectWriter.WriteEndElement()
```
aholstrup1, line 151: Consider either requiring a .nuspec file as input or construcitng a nuspec and using the nuget.exe to create the package from the nuspec.

aholstrup1 (approved):

NuGet/Get-BcNuGetPackage.ps1
```diff
@@ -1,101 +1,133 @@
...
+            Dump "::group::Search NuGetFeed $($feed.Url)"
             try {
-                $package = Invoke-RestMethod -UseBasicParsing -Method GET -Headers $headers -Uri $packageVersion.'@id'
-                $contentUrl = $package.packageContent
+                try {
+                    $nuGetFeed = [NuGetFeed]::Create($feed.Url, $feed.Token, $feed.Patterns, $silent)
+                }
+                catch {
+                    Dump "Initiation of NuGetFeed failed. Error was $($_.Exception.Message)"
+                    continue
```
mazhelez, line 69: Why `continue` here?

mazhelez (dismissed):

NuGet/New-BcNuGetPackage.ps1
freddydk: A general thing in ContainerHelper is to have as few dependencies as possible on external tools that needs to be installed.

NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -0,0 +1,194 @@
...
+  NuGet Token for authenticated access to the NuGet Server
+  If not specified, the NuGet Server is accessed anonymously (and needs to support this)
+ .PARAMETER packageName
+  Package Name to search for.
+  This can be the full name or a partial name with wildcards.
+  If more than one package is found, matching the name, an error is thrown.
+ .PARAMETER version
+  Package Version, following the nuget versioning rules
+  https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges
+ .PARAMETER silent
```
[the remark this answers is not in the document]
freddydk: I think it is a good idea to to use -verbose - I can however not use Write-Verbose as grouping f.ex. will not work:

VERBOSE: ::group::NUSPEC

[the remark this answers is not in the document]
freddydk: will fix this

NuGet/Get-BcNuGetPackage.ps1
freddydk: Idea was that if you have 3 trusted feeds and one feed doesn't work - it continues to search for the package in other feeds instead of failing hard.
But, I actually think it should fail instead - will do that.

NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -0,0 +1,194 @@
...
+        [string] $version = '0.0.0.0',
+        [switch] $silent,
+        [Parameter(Mandatory=$false)]
+        [ValidateSet('Earliest','Latest','Exact','Any')]
+        [string] $select = 'Latest',
+        [Parameter(Mandatory=$true)]
+        [alias('appSymbolsFolder')]
+        [string] $folder,
+        [Parameter(Mandatory=$false)]
+        [string] $copyInstalledAppsToFolder = "",
```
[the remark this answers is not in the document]
freddydk: It could - the copyInstalledAppsToFolder is however used already outside of BcContainerHelper - so I won't change this now.

DominicSaladin:
I'm not sure if I'm allowed to comment on those pull requests, but I just looked at the implementation, because I attended the webinar of Kamir yesterday.
It looks like the GetVersions function of the NuGetFeed class returns all versions including prereleases, but powershell can't parse those version numbers.

For example this prerelease version number: `2023.8.20231214.55563-alpha`
This results in "ForEach-Object: The value `2023.8.20231214.55563-alpha" cannot be converted to the type "System.Version". Error: "The input string was in the incorrect format.`

Maybe you need to filter those out.

freddydk:
> I'm not sure if I'm allowed to comment on those pull requests, but I just looked at the implementation, because I attended the webinar of Kamir yesterday. It looks like the GetVersions function of the NuGetFeed class returns all versions including prereleases, but powershell can't parse those version numbers.
> 
> For example this prerelease version number: `2023.8.20231214.55563-alpha` This results in "ForEach-Object: The value `2023.8.20231214.55563-alpha" cannot be converted to the type "System.Version". Error: "The input string was in the incorrect format.`
> 
> Maybe you need to filter those out.

We have been debating how to handle pre-release.
According to the nuget standard - pre-release is only part of a semver compatible versioning - i.e. 3 digits - and in the current implementation I always use 4 tuples (System.Version compatible) - on which I don't think nuget supports adding a pre-release tag - or????

mazhelez (dismissed):

freddydk:
It looks like Azure DevOps supports a pre-release tag after a 4 tuple version number:
image

DominicSaladin:
> It looks like Azure DevOps supports a pre-release tag after a 4 tuple version number: image

Yes, Azure does support it and that is the problem. I don't think that the System.Version supports anything similar.

freddydk:
@DominicSaladin - I have added prerelease support.
Get-BcNuGetPackage and Download-BcNuGetPackage now has a flag called -allowPrerelease, with includes prerelease builds in the search. Without this flag - prerelease builds are ignored in the search
New-BcNuGetPackage has a prereleaseTag parameter in order to create a prerelease package.

DominicSaladin:
@freddydk Yeah, that worked perfectly!
Got another question for you. Wouldn't it make sense to add a output path parameter? Inside the output path you could create a folder with all the content you just downloaded for the selected nuget package?

> I am still missing the piece, which allows you to download a package for Business Central version 21.0, which should skip the versions published for 22 or 23.
That is going to be difficult. In my example the version is `2023.8.0.0` so you have no idea for which BC version it is. And I know that the standard NuGet api has no such filtering options :/

freddydk:
> That is going to be difficult. In my example the version is `2023.8.0.0` so you have no idea for which BC version it is. And I know that the standard NuGet api has no such filtering options :/
Correct - there is no such filtering mechanism - but I can look at the dependencies in the nuspec - if they point to the wrong Application - I can grab another one package.
Today - an error is thrown.

aholstrup1 (approved):

PR #3278, opened 2023-12-20
Pass useDevEndpoint parameter when importing test toolkit during run-alpipeline

aholstrup1:


AppHandling/Run-AlPipeline.ps1
```diff
@@ -1320,6 +1320,12 @@ Measure-Command {
             "environment" = $environment
         }
     }
+    if ($useDevEndpoint) {
+        $Parameters += @{
+            "devEndpoint" = $useDevEndpoint
+            "credential" = $credential
```
freddydk, line 1326: Credential should probably only be passed if set (or if not bcAuthContext)

aholstrup1: Doesn't this set the credential if it isn't provided as an input? https://github.com/microsoft/navcontainerhelper/blob/d24fd0fa773c32a7d45b8d1354cc5c5121733bf5/AppHandling/Run-AlPipeline.ps1#L660-L667

freddydk: Correct - if we are using a container, then yes, but if BcAuthContext is provided, it uses an online environment for the publish - which doesn't need a credentials object. It probably doesn't fail if provided, but maybe just set useDevEndPoint in the if above and then elseif (useDevEndpoint) here ?

freddydk (dismissed):

freddydk (approved):

PR #3297, opened 2024-01-18
Closes #3296

### LINKED ISSUE #3296: Compile App With BC Compiler Folder Override
It would be nice if there was an override to this function in Run-AlPipeline.

image

Will submit a PR shortly.

MattTraxinger:
Closes #3296

Adding an override to Compile-AppWithBcCompilerFolder. There is one for Compile-AppInBcContainer already, but I think this one should exist too.


freddydk:
Thanks Matt - LGTM

PR #3329, opened 2024-02-05
fix: escape bakfile environment variable

warlof:
While using a custom bak file containing special characters (space and parenthesis), I fall on an exception due to the fact bak path was not escaped.

This fix apply the same encapsulate than the one done for licence file.


freddydk (approved):

freddydk:
Thanks @warlof - merged and will be included in next preview

PR #3341, opened 2024-02-16
JIT install ps7

freddydk:
BC 24.0 will require PS7.4.1 to be installed in containers

### DIFF
--- a/ContainerHandling/New-NavContainer.ps1
+++ b/ContainerHandling/New-NavContainer.ps1
@@ -1745,10 +1745,12 @@ if (Test-Path "c:\run\my\dotnet-win.exe") { Write-Host "Installing dotnet 6.0.13
               . (Join-Path $runPath $MyInvocation.MyCommand.Name)
             ') | Set-Content -Path "$myfolder\HelperFunctions.ps1"
         }
-        Write-Host "Patching container to install dotnet 8.0.0"
+        Write-Host "Patching container to install dotnet 8.0.0 and PowerShell 7.4.1"
         Download-File -source "https://download.visualstudio.microsoft.com/download/pr/2a7ae819-fbc4-4611-a1ba-f3b072d4ea25/32f3b931550f7b315d9827d564202eeb/dotnet-hosting-8.0.0-win.exe" -destinationFile (Join-Path $myFolder "dotnet-win8.exe")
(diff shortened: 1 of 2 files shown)


Koubek:
@freddydk Can we expect the same changes common for all versions (keeping one generic image for all artifact/major versions)? I believe yes but I prefer to ask and get your confirmation :)

freddydk:
> @freddydk Can we expect the same changes common for all versions (keeping one generic image for all artifact/major versions)? I believe yes but I prefer to ask and get your confirmation :)

I hope so

freddydk:
> > @freddydk Can we expect the same changes common for all versions (keeping one generic image for all artifact/major versions)? I believe yes but I prefer to ask and get your confirmation :)
> 
> I hope so

I will however also try to create a "BC 24+ only" image, without everything needed for the older versions - this might be much slimmer and more efficient if we can avoid .net framework alltogether.

mazhelez (approved):

Koubek:
Splitting up isn't perfect as it adds some level of complexity, especially for those who need to adjust the base images and also for multiple OS versions. But on the other hand, I can understand the idea of using slimmer images is sexy :) Anyway, I believe most of the partners will need both for a certain amount of time so far so I would rather wait with splitting e.g. at least until the next BC release (v25) :)

freddydk:
The goal is that the "bloated"/"current" image will be able to run BC24 as well - and then there could be a slimmer version for BC24+

Koubek:
Amazing 👍

freddydk:
FYI - The new generic image with dotnet 8.0.2 is 8.02GB :-) (based on dotnet/framework/runtime image, which is 4.55GB = 3½GB to download)
a 24+ only version of the generic image is 6.32GB (based on windows/servercore image, which is 4.29GB = 2GB to download)

Koubek:
Can I understand the images are still not available publicly? I can see 1.0.2.14 as the latest one probably. But I have searched just here and maybe I have overlooked the new one(s).

freddydk:
> Can I understand the images are still not available publicly? I can see 1.0.2.14 as the latest one probably. But I have searched just [here](https://mcr.microsoft.com/v2/businesscentral/tags/list) and maybe I have overlooked the new one(s).

Working on it

freddydk:
Generic image with tag 1.0.2.15 has been released
Note that ASP.NET Core 1.1 is no longer on the image, meaning that for BC versions 15-18 - you will have to JIT install ASP.NET Core 1.1 (which BcContainerHelper is doing) - or use 1.0.2.14

Issue #3339, opened 2024-02-15
New-AadAppsForBC - The property 'Id' cannot be found on this object.


DBiernat:
**Describe the issue**

Using New-AadAppsForBc ends up in the following error Message: The property 'Id' cannot be found on this object.

**Scripts used to the issue**
```
New-AadAppsForBC -appIdUri $idUri -bcAuthContext $authContext
```
**Full output of scripts**
```
Creating AAD App for WebClient
The property 'Id' cannot be found on this object. Verify that the property exists.
At C:\Program Files\WindowsPowerShell\Modules\BcContainerHelper\6.0.5\AzureAD\New-AadAppsForBc.ps1:169 char:5
+     $admspwd = Add-MgApplicationPassword -ApplicationId $SsoAdApp.Id  ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : PropertyNotFoundStrict
```

**Additional context**
There is a typo in Line 169: "$**S**soAdApp.Id" has to be "$**s**soAdApp.Id"

freddydk:
Variable names in PowerShell are case insensitive, so while you are right that this is a type, it isn't the reason for the problem.
Just tried this:

```
$bcAuthContext = New-BcAuthContext -includeDeviceLogin -scopes "https://graph.microsoft.com/.default"
$AdProperties = New-AadAppsForBc -appIdUri https://mycontainer.directionsemea2019demo.onmicrosoft.com -bcAuthContext $bcAuthContext
```

Logged in as my test account on tenant directionsemea2019demo.onmicrosoft.com and this works fine:

Not sure what is wrong though - typically you would get another error first, like:
![image](https://github.com/microsoft/navcontainerhelper/assets/10775043/8ff81a88-4113-4d26-9b7d-960d306670cf)

(and if erroractionpreference is stop - you would never get the second)

DBiernat:
To be honest, it was new to me, that variable names are Case Insensitive. 😅

The first difference is, that I'm using an App rather than the Device Login. 

```
$authContext = New-BcAuthContext -tenantID $tenantID -clientID $clientID -scopes "https://graph.microsoft.com/.default" -clientSecret $clientSecret
```

[...]

PR #3342, opened 2024-02-19
Changed local variable casing

ronaldnl:
As described in issue [3339](https://github.com/microsoft/navcontainerhelper/issues/3339).
The casing causes the command to fail.

I changed both local vars to start on all places with a non capital to solve the issue which will make the command fail


freddydk:
I don't mind including this - but please then also change the `.id` in line 193 to `.Id` - which actually could be a problem, since fields in .net object are case sensitive.
You would still need to find out why your machine sees everything as case sensitive though:-)

freddydk (approved):

PR #3375, opened 2024-02-27
Disallow invoking Run-ALPipeline with keepContainer without specifying credential

mazhelez:
If `keepContainer` is specified, `credential` must also be specified, as otherwise the container will be created with a random password and there will be no way to access it.


freddydk:
Please also remove lines 670 - 678 (the dump on whether credentials is specified) - there is no reason to dump this info anymore.

AppHandling/Run-AlPipeline.ps1
```diff
@@ -485,6 +485,11 @@ if ($sharedFolder -and !(Test-Path $sharedFolder -PathType Container)) {
     throw "If sharedFolder is specified, it must be an existing folder"
 }
 
+if($keepContainer -and !$credential) {
```
freddydk, line 488: Please also remove lines 670 - 678 (the dump on whether credentials is specified) - there is no reason to dump this info anymore.

freddydk (changes requested):

freddydk (changes requested):
Please also remove lines 670 - 678 (the dump on whether credentials is specified) - there is no reason to dump this info anymore.

AppHandling/Run-AlPipeline.ps1
mazhelez: Done.

freddydk (approved):

Issue #3378, opened 2024-02-28
"Get-WindowsOptionalFeature: Class not registered" on W11 with PowerShell 7.4.1


jwikman:
**Describe the issue**
I'm trying to create a new container, from VSCode running as an administrator (yes, I know I can do that while not running as Admin, but I got other permissions issues that I didn't have time to dive into right now).

In `New-BcContainer` there is a call to `Get-HypervState` if process is running as Admin.
In that function, the state of Hyper-V is checked via `Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V -Online`
But on my W11 machine, running PowerShell 7.4.1, that results in the error (after a loooong time)
`Get-WindowsOptionalFeature: Class not registered
`

```
>$PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.1
PSEdition                      Core
GitCommitId                    7.4.1
OS                             Microsoft Windows 10.0.22631
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

**Scripts used to create container and cause the issue**
Anything that calls `New-BcContainer` in an elevated mode.

**Full output of scripts**
Below is after I removed `Write-Host "HyperV is $(Get-HypervState)"` from `New-BcContainer`, otherwise I got stuck after the second line...
```
BcContainerHelper is version 6.0.7-preview1150
BcContainerHelper is running as administrator
UsePsSession is True
Host is Microsoft Windows 11 Enterprise - 10.0.22631.3155
Docker Client Version is 24.0.7
Docker Server Version is 24.0.7
```

**Additional context**
- does it happen all the time?
  - Always if running as admin (unless uncomment the `Write-Host` line mentioned above 😉)
- did it use to work?
  -  Yes, before I updated to PowerShell 7

freddydk:
Strange
![image](https://github.com/microsoft/navcontainerhelper/assets/10775043/a533623e-8830-46ff-8764-6116ebf229d0)

Do you have Hyper-V installed?

jwikman:
Yes, strange indeed.

It seems as if this issue is from `Get-WindowsOptionalFeature`, and not related to BcContainerHelper.
See https://github.com/PowerShell/PowerShell/issues/13866 for more info.

Even this fails: 
```
PowerShell 7.4.1
[...]

PR #3382, opened 2024-03-01
Resiliency for Get-WindowsOptionalFeature

jwikman:
Solves #3378 by catching the specific error and continue.

This is the output of `New-BcContainer` when running as administrator and having the mentioned issue with `Get-WindowsOptionalFeature -online`:
```
BcContainerHelper is running as administrator
Cannot check Hyper-V status.
HyperV is
```
At least we do not stop anymore with this.


freddydk (approved):

freddydk:
FYI, if you use Fixes #3378 in your description (instead of Solves) - then GitHub will automagically close the issue when merged

PR #3388, opened 2024-03-05
add links in manifest to enable direct links in psgallery

ChrisBlankDe:


freddydk (changes requested):
See comments

ChrisBlankDe:
> The .psd1 files are autogenerated from these lines:
> 
> https://github.com/microsoft/navcontainerhelper/blob/5eef4ac0d52d40151cba7f5fc0e97c8a3783b4df/.github/workflows/CI.yaml#L193
> 
> and
> 
> https://github.com/microsoft/navcontainerhelper/blob/5eef4ac0d52d40151cba7f5fc0e97c8a3783b4df/.github/workflows/Release.yaml#L63
> 
> -LicenseUri and -ProjectUri
> 


The maifest ist not autogenerated but auto-updated.
As we do not overwrite the values in Update-ModuleManifest, the values already set in the psd1 will be retained. (Just Tested)

Should I add it to the actions anyway?

freddydk:
> > The .psd1 files are autogenerated from these lines:
> > https://github.com/microsoft/navcontainerhelper/blob/5eef4ac0d52d40151cba7f5fc0e97c8a3783b4df/.github/workflows/CI.yaml#L193
> > 
> > and
> > https://github.com/microsoft/navcontainerhelper/blob/5eef4ac0d52d40151cba7f5fc0e97c8a3783b4df/.github/workflows/Release.yaml#L63
> > 
> > -LicenseUri and -ProjectUri
> 
> The maifest ist not autogenerated but auto-updated. As we do not overwrite the values in Update-ModuleManifest, the values already set in the psd1 will be retained. (Just Tested)
> 
> Should I add it to the actions anyway?

Yes, please do - I don't want to have anything in the .psd1 files which needs to be maintained.

BcContainerHelper.psd1
```diff
@@ -266,10 +266,10 @@ PrivateData = @{
         # Tags = @()
 
         # A URL to the license for this module.
-        # LicenseUri = ''
```
freddydk, line 269: please remove the changes to the .psd1 files (the other ones have become binary)
image

freddydk (changes requested):

freddydk (approved):

PR #3401, opened 2024-03-08
Add Schedule parameter to Publish-PerTenantExtensionApps

M4fin:
suggestion to add a schedule parameter to allow deployment of the app to the next major version via CI/CD Pipeline


freddydk:
Could you add the documentation to the API description  on which the PR is based?

M4fin:
Sure, documentation is [here](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/administration/resources/dynamics_extensionupload#properties)

Saas/Publish-PerTenantExtensionApps.ps1
```diff
@@ -42,6 +42,8 @@ function Publish-PerTenantExtensionApps {
         $appFiles,
         [ValidateSet('Add','Force')]
         [string] $schemaSyncMode = 'Add',
+        [ValidateSet('Current','Next minor','Next major')]
+        [string] $schedule,
```
freddydk, line 46: Please modify the values to be the same as the supported values in schedule and then only set body.schedule if schedule is not blank.
Also we need to support not setting schedule at all - maybe allow blank and have default be blank.

freddydk (changes requested):

Saas/Publish-PerTenantExtensionApps.ps1
M4fin: changed in the last commit

Saas/Publish-PerTenantExtensionApps.ps1
```diff
@@ -42,6 +42,8 @@ function Publish-PerTenantExtensionApps {
         $appFiles,
         [ValidateSet('Add','Force')]
         [string] $schemaSyncMode = 'Add',
+        [ValidateSet(' ','Current version','Next minor version','Next major version')]
```
freddydk, line 45: Please use empty string as default instead of a space.

freddydk (changes requested):
And please add a line to the top section in releasenotes.txt, thanks

Saas/Publish-PerTenantExtensionApps.ps1
M4fin: Fixed

M4fin:

ReleaseNotes.txt
```diff
@@ -16,6 +16,7 @@ If environment is specified as a Web Client URL, and BcAuthContext contains user
 Including caching of appinfos in CompilerFolder cache (to save time when caching on GitHub Actions)
 
 6.0.19
+Add Schedule parameter to Publish-PerTenantExtensionApps
```
freddydk, line 19: Please move the release note to 6.0.21

freddydk (changes requested):

freddydk (approved):

Saas/Publish-PerTenantExtensionApps.ps1
```diff
@@ -42,6 +42,8 @@ function Publish-PerTenantExtensionApps {
         $appFiles,
         [ValidateSet('Add','Force')]
         [string] $schemaSyncMode = 'Add',
+        [ValidateSet('','Current version','Next minor version','Next major version')]
```
mazhelez, line 45: No need for `''`, it's not a mandatory parameter.

mazhelez, line 45:
```suggestion
        [ValidateSet('Current version','Next minor version','Next major version')]

```

Saas/Publish-PerTenantExtensionApps.ps1
```diff
@@ -42,6 +42,8 @@ function Publish-PerTenantExtensionApps {
         $appFiles,
         [ValidateSet('Add','Force')]
         [string] $schemaSyncMode = 'Add',
+        [ValidateSet('','Current version','Next minor version','Next major version')]
+        [string] $schedule = '',
```
mazhelez, line 46:
```suggestion
        [string] $schedule,

```

mazhelez (approved):

Saas/Publish-PerTenantExtensionApps.ps1
freddydk: most pipelines will have code where the value of this comes from a variable, which comes from a file or like. And this:

```
function test {
    Param(
        [ValidateSet('Current version','Next minor version','Next major version')]
        [string] $schedule
    )

    Write-Host "'$test'"
}

$schedule = ''
test -schedule $schedule
```

throws an error:
test : Cannot validate argument on parameter 'schedule'. The argument "" does not belong to the set "Current version,Next minor version,Next major version" specified by the ValidateSet attribute. Supply an 
argument that is in the set and then try the command again.

Which means that people will have to generate two code-paths for this scenario - and we do not want that.

PR #3430, opened 2024-03-19
improve output to match build system

ChrisBlankDe:
github actions and azure pipelines use different environment variables we could use to identify the running system.
we should use this possibility to output formatting commands for the given system.


freddydk:
I don't think testing for various environment variables deep inside ContainerHelper is the right approach.
I would accept a PR with 2 new properties in ContainerHelper config called IsGitHubActions and IsAzureDevOps - both default to $false. Then I would have two functions in HelperFunctions called WriteGroupStart and WriteGroupEnd with a message parameter, which would check for the two settings.
also - all places in Run-AlPipeline, where a group start is called - this function should be used and the value of githubActions in Run-AlPipeline should be defaulted to the setting - and the value of the parameter should be written into the bccontainerhelperconfig.IsGitHubActions (for that session).

ChrisBlankDe:
i wanted to change as little as possible, hence the somewhat inelegant solution. your approach is of course better.

but why shouldn't we automatically recognize the build system when loading the bccontainerhelper (possibly overridable by a setting)? if it only works nicely when the user activates a setting, most people will overlook it and not have the best experience.

BC.HelperFunctions.ps1
```diff
@@ -120,6 +120,9 @@ function Get-ContainerHelperConfig {
             "MSAppsNuGetFeedUrl" = 'https://dynamicssmb2.pkgs.visualstudio.com/DynamicsBCPublicFeeds/_packaging/MSApps/nuget/v3/index.json'
             "TrustedNuGetFeeds" = @(
             )
+            "IsGitHubActions" = [Boolean]$env:GITHUB_ACTIONS
```
freddydk, line 123: In PowerShell - [boolean]"false" is also true - I would prefer 

"IsGitHubActions" = ($env:GITHUB_ACTIONS -eq "true")

(and the same with the other)

HelperFunctions.ps1
```diff
@@ -1299,3 +1299,18 @@ function ReplaceCDN {
     }
     $sourceUrl
 }
+
+function Write-GroupStart([string] $Message) {
+    switch ($true) {
+        $bcContainerHelperConfig.IsAzureDevOps { Write-Host "##[group]$Message"; break }
+        $bcContainerHelperConfig.IsGitHubActions { Write-Host "::group::$Message"; break }
```
freddydk, line 1306: Please change the order of these to have IsGitHub on top and test for IsAzureDevOps afterwards.

I can see situations where TF_BUILD could be set to something in GitHub - something a partner could define themselves. I don't see anybody defining an environment variable called GITHUB_ACTIONS themselves.

freddydk (changes requested):
A few changes and then we can merge this one.

freddydk:
@ChrisBlankDe - please fix the remaining two issues and let's get this in

BC.HelperFunctions.ps1
ChrisBlankDe: fixed

HelperFunctions.ps1
ChrisBlankDe: fixed

freddydk (approved):

freddydk:
Thanks :-)

PR #3434, opened 2024-03-19
Create AppInfo cache on creating a new compiler folder

ChrisBlankDe:
the appinfo cache could be created right after the creation of the compiler folder.

this improves build performance if the compile folder is created and cached without running a compile before caching.

In our case we create a bunch of compile folders for different versions and localizations and then create an vm image which is used for build agents


freddydk (approved):
LGTM

PR #3439, opened 2024-03-23
Enable limited features of BcContainerHelper on macOS

ashirokikh:
With AL Language support on macOS, some of the BcConatinerHelper features can be handy.

This is to enable the module import for macOS and an AL app compilation without a container. The following commands have been confirmed working:

- **Get-BCArtifactUrl** - returns the URL as expected
- **Download-Artifacts** - saves the artifacts into the bcartifacts.cache folder (at /Users/$myUsername/.bcartifacts.cache)
- **New-BcCompilerFolder** - creates a compiler folder (no issues with latest vsix file with altool)
- **Compile-AppWithBcCompilerFolder** - compiles an AL app using the created compiler folder

Tested on OS versions:
Windows 10, PowerShell 5.1
macOS 14.4, pwsh 7.4
Ubuntu 20.04 (WSL2), pwsh 7.4

image


ashirokikh:

freddydk:
Before looking at this, I have to ask why?

Why would you want to run DevOps on MacOS, which is 10 times the price of Linux runners and less performant?
image

What is the purpose of running Get-BcArtifactUrl on MacOS when you really cannot use the URL for much?

Remember that enabling stuff like this comes with a support cost, which in the end lands with me/Microsoft - so there need to be a good reason.

The reason for Linux is that it is ½ the price of windows for DevOps and faster.

ashirokikh:
Thank you for looking into this, those are very good points.

The main goal for this is to further enable macOS as an AL development environment, not as a pipeline agent host OS. I know BcContainerHelper is supposed to help with Docker but it's still a very powerful tool besides that. For instance, getting the BC artifacts function is especially helpful to navigate and get familiar with the base app source code and compare several versions to make a solution compatible with them. The Auth scripts can be used to refresh tokens without using any http client. 

Personally, as part of a web/mobile development team working on macOS, I still work on BC backend developing in AL, and getting artifacts is the first thing that I wanted to do to carry on after switching the platform. I saw some questions from other macOS users about BcContainerHelper here and there are already some BC developers working on macs, hence I thought this change could be helpful. Although many functions may not be working, the ones that are can still be very helpful. For instance, with this change, an AL app can be compiled against several BC versions using several compiler folders without a container.

Looking broader, according to the [Stack Overflow Developer Survey](https://survey.stackoverflow.co/2023/#section-most-popular-technologies-operating-system), macOS market share across professionals is 33%. Enabling this power tool for the platform may attract more developers to take up a BC developer role without having to switch the preferred and familiar work environment.

There may be not much else we can do with the downloaded artifacts and platform today on macOS. Yet with the trend of going cross-platform with .NET and the lower price of running Linux compute resources in the cloud, the chances are high that in future we may see running BC in a Linux container, and macOS is capable of that. Until then, enabling the tool for macOS can interfere with the current priorities, I understand. 

If this is currently not going bring enough value, please close this pull request. Thank you again!

HelperFunctions.ps1
```diff
@@ -1276,6 +1282,11 @@ function RunAlTool {
     if ($isLinux) {
         $alToolExe = Join-Path $path 'extension/bin/linux/altool'
         & /usr/bin/env sudo pwsh -command "& chmod +x $alToolExe"
+    } 
+    elseif ($isMacOS) {
+        $alToolExe = Join-Path $path 'extension/bin/darwin/altool'
+        Write-Host "Setting execute permissions on altool"
+        & chmod +x $alToolExe
```
freddydk, line 1289: No sudo needed?

CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -254,7 +267,7 @@ try {
                     $newAlcConfig | ConvertTo-Json | Set-Content -Path $alcConfigPath -Encoding utf8NoBOM
                 }
             }
-        }
+        }
```
freddydk, line 270: Remove spaces

HelperFunctions.ps1
```diff
@@ -1058,6 +1058,12 @@ function GetAppInfo {
         $alToolExe = Join-Path $alcPath 'altool'
         & /usr/bin/env sudo pwsh -command "& chmod +x $alToolExe"
     }
+    elseif ($isMacOS) {
+        $alcPath = Join-Path $binPath 'darwin'
+        $alToolExe = Join-Path $alcPath 'altool'
+        Write-Host "Setting execute permissions on altool"
+        & chmod +x $alToolExe
```
freddydk, line 1065: No sudo needed?

freddydk (commented):
I can see that it is fairly close to linux, a few questions

HelperFunctions.ps1
ashirokikh: Correct, worked without it since it's a user repository.

HelperFunctions.ps1
ashirokikh: The current user has permissions to the user's folder, no admin permissions required.

CompilerFolderHandling/New-BcCompilerFolder.ps1
ashirokikh: Removed

freddydk (approved):

PR #3470, opened 2024-04-05
Set timeouts to zero for Backup-SqlDatabase

glavanovic:
During imports of large databases a timeout exception error occurs. With this change timeouts will be set to zero to run as long as neccessary.


Bacpac/Backup-NavContainerDatabases.ps1
```diff
@@ -68,7 +68,12 @@ try {
             $params = @{}
             if ($compress) { $params += @{ "CompressionOption" = "On" } }
             if ($databaseCredential) { $params += @{ "credential" = $databaseCredential } }
-            Backup-SqlDatabase -ServerInstance $serverInstance -database $database -BackupFile $bakFile @params
+
```
freddydk, line 71: Text says importing?
Is this really a problem with backup? Database size in SQL Express is limited - did you ever run into this?

PR #3496, opened 2024-04-20
Use sessions

freddydk:
Add PowerShell configurations and PSRemoting when creating containers
Allow standard Container sessions (when host is running PS5 in admin mode) to be to PS5 or PS7 (base on version and usePwshForBC24 flag)
Use WinRm to create a session to a container when running PS7 or non-admin mode.
WinRm session uses HTTPS unless $bccontainerHelperConfig.useSslForWinRm is set to $false
Using HTTP for WinRm requires you to add trustedHosts to the hosts winrm configuration (if not running as administrator)

### DIFF
--- a/ContainerHandling/Get-NavContainerSession.ps1
+++ b/ContainerHandling/Get-NavContainerSession.ps1
@@ -17,6 +17,7 @@ function Get-BcContainerSession {
     Param (
         [string] $containerName = $bcContainerHelperConfig.defaultContainerName,
         [switch] $tryWinRmSession = $bccontainerHelperConfig.tryWinRmSession,
+        [switch] $usePwsh = $bccontainerHelperConfig.usePwshForBc24,
         [switch] $silent,
         [switch] $reinit
     )
@@ -37,13 +38,21 @@ function Get-BcContainerSession {
             }
         }
         if (!$session) {
+            [System.Version]$platformVersion = Get-BcContainerPlatformVersion -containerOrImageName $containerName
+            if ($platformVersion -lt [System.Version]"24.0.0.0") {
+                $usePwsh = $false
+            }
+            $configurationName = 'Microsoft.PowerShell'
+            if ($usePwsh) {
+                $configurationName = 'PowerShell.7'
+            }
             if ($isInsideContainer) {
                 $session = New-PSSession -Credential $bcContainerHelperConfig.WinRmCredentials -ComputerName $containerName -Authentication Basic -UseSSL -SessionOption (New-PSSessionOption -SkipCACheck -SkipCNCheck)
             }
             elseif ($isAdministrator) {
                 try {
                     $containerId = Get-BcContainerId -containerName $containerName
-                    $session = New-PSSession -ContainerId $containerId -RunAsAdministrator -ErrorAction SilentlyContinue
+                    $session = New-PSSession -ContainerId $containerId -RunAsAdministrator -ErrorAction SilentlyContinue -ConfigurationName $configurationName
                 }
                 catch {}
             }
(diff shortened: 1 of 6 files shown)


ContainerHandling/Get-NavContainerSession.ps1
```diff
@@ -37,34 +39,39 @@ function Get-BcContainerSession {
             }
         }
         if (!$session) {
+            [System.Version]$platformVersion = Get-BcContainerPlatformVersion -containerOrImageName $containerName
+            if ($platformVersion -lt [System.Version]"24.0.0.0") {
+                $usePwsh = $false
```
aholstrup1, line 44: Why not respect the $bccontainerHelperConfig.usePwshForBc24 setting?

ContainerHandling/Get-NavContainerSession.ps1
```diff
@@ -37,34 +39,39 @@ function Get-BcContainerSession {
...
-                    $session = New-PSSession -ContainerId $containerId -RunAsAdministrator -ErrorAction SilentlyContinue
+                    $session = New-PSSession -ContainerId $containerId -RunAsAdministrator -ErrorAction SilentlyContinue -ConfigurationName $configurationName
                 }
                 catch {}
             }
             if (!$session) {
-                if (!$tryWinRmSession) {
-                    throw "Unable to create a session for container $containerName (tryWinRmSession is false)"
+                if (!($alwaysUseWinRmSession -or $tryWinRmSession)) {
+                    throw "Unable to create  session for container $containerName (alwaysUseWinRmSession and tryWinRmSession are both false)"
```
aholstrup1, line 62:
```suggestion
                    throw "Unable to create session for container $containerName (alwaysUseWinRmSession and tryWinRmSession are both false)"

```

aholstrup1 (dismissed):

aholstrup1 (approved):

StefanMaron:
May I ask what the benefit of using the WinRM is? Seems like I had the service disabled on my local runner VM and thus, all my pipeline runs over the weekend failed. I will just enable that service again to fix it, but I am still curious what benefit this brings.

freddydk:
It shouldn't need to be enabled on the host - but you say it is?
You can open a session to a container in PS5 running as admin without winrm.
If you are running as non-admin or you are running PS7 - you need winrm.
I will modify the implementation to not touch winrm if you are running PS5 in admin mode - meaning that everything will work as before - will be in a new version in a few hours.

StefanMaron:
Seems like my runner uses PS7 so that explains it I guess.
I have no problem to have the service enabled, I was just curious why it is a requirement now, and what benefit that adds :)
image

freddydk:
I will also fix it so that it won't give an error - but instead just use docker exec (which is much slower though)

ContainerHandling/Remove-NavContainer.ps1
```diff
@@ -62,6 +62,18 @@ try {
             . (Join-Path $PSScriptRoot "updatehosts.ps1") -hostsFile "c:\windows\system32\drivers\etc\hosts" -theHostname $tenantHostname -theIpAddress ""
         }
 
+        if ($isAdministrator) {
+            try {
+                [xml]$conf = winrm get winrm/config/client -format:pretty
+                $trustedHosts = $conf.Client.TrustedHosts.Split(',')
+                if ($trustedHosts -contains $containerName) {
+                    Write-Host "Removing $containerName from trusted hosts ($($trustedHosts -join ','))"
+                    winrm set winrm/config/client "@{TrustedHosts=""$(@($trustedHosts | Where-Object { $_ -ne $containerName }) -join ',')""}" | Out-Null
```
StefanMaron, line 71: @freddydk I think this was the place where it threw and error in my pipeline. Is the host name configuration even needed in pipelines?

freddydk:
Yep - I got it - it cannot read client config if the service isn't running

PR #3539, opened 2024-05-15
Fix infinite loop due special character in filename

RemkoD:
Today, I encountered an infinite loop when creating a new container that required a new docker image. The repeated message in the shell indicated the problem: 

`Waiting for c:\bcartifacts.cache\deaz3r2n.5od\NAVDVD\ModernDev\program files\Microsoft Dynamics NAV\230\AL Development Environment\ALLanguage\[Content_Types].xml to be available`

The file `[Content_Types].xml` contains special characters, causing Test-Path to return $false. By specifying that the path is a literal path, the function returns $true.

### DIFF
--- a/ContainerHandling/New-NavImage.ps1
+++ b/ContainerHandling/New-NavImage.ps1
@@ -78,7 +78,7 @@ function RoboCopyFiles {
         RoboCopy "$source" "$destination" "$files" /e /NFL /NDL /NJH /NJS /nc /ns /np /mt /z /nooffload | Out-Null
         Get-ChildItem -Path $source -Filter $files -Recurse | ForEach-Object {
             $destPath = Join-Path $destination $_.FullName.Substring($source.Length)
-            while (!(Test-Path $destPath)) {
+            while (!(Test-Path -literalPath $destPath)) {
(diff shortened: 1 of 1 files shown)


freddydk:
The fix might be fine - will check this - but how did that file appear in that folder?
It looks like this fix removes a symptom instead of the cause.

The folder normally contains:

image

Meaning that there wouldn't be a problem.
Did you or I unpack the .vsix file in that folder?
I would like to understand how this happened before merging this as there might be other problems.

If you can use Flush-ContainerHelperCache and then run a script, which ends up with this problem - then I will see if this indeed is the right fix.

Thanks

freddydk:
I looked through the code and cannot find anywhere where I unpack the .vsix inside the artifacts cache - I think I always unpack it into a temp folder

RemkoD:
@freddydk, You are most likely correct. I was creating a container with the vsixFile parameter, assuming this caused the unpacking of the vsix file. But I also remember extracting the vsix file recently, and I might have forgotten to remove the folder afterward. 

I will try it again later today with a clean cache!

RemkoD:
It took a bit longer than expected, but I can confirm that with a clean cache the error doesn't occur.

PR #3615, opened 2024-08-19
fix read environment variable when backup or restoring db

lukas-pfeifhofer:
In some situations the reading of environment variables from the container not works as expected with the `Invoke-ScriptInBcContainer` cmdlt, f.i. the latest online sandbox (version 24.4.22295.23050) for Italy on a self-hosted agent not returns any value, on an on-prem sandbox (version 23 and also version 24) instead returns a value as expected.

The following statement was launched on pwsh 7.4.2 with docker version 20.10.21, but returns no value:

```pwsh
Invoke-ScriptInBcContainer -containerName $containerName -scriptblock { $env:IsBcSandbox }
```

Using the following fix for this issue works as expected, and is much more faster then invoking the script:

```pwsh
$inspect = docker inspect $containerName | ConvertFrom-Json
$inspect.Config.Env | Where-Object { $_ -eq "IsBcSandbox=Y" }
```


freddydk:
Thanks... - will have a look at this tomorrow

freddydk:
The fix here is good - but your finding actually points to a problem, which I don't really know the consequence off.
When creating a PSSession or a WinRm session to a container, the environment variables specified when creating the container are not accessible:

image

I am pretty sure that it hasn't been like this all the time - else the code that you replace in this PR would never have worked.

I will approve this PR and investigate if this finding has other impacts...

freddydk (approved):

PR #3626, opened 2024-09-03
Get-NavContainers.ps1 : Adding a parameter: ContainerName to filter on the container name

JFLammech:
Update Get-NavContainers.ps1
Adding a parameter: ContainerName
to filter on the container name


JFLammech:

freddydk:
Why not just use:

```
Get-BcContainers | Where-Object { $_ -like 'bc*' }
```

freddydk:
With -includeLabels - it would be

```
Get-BcContainers -includeLabels | Where-Object { $_.name -like 'bc*' }
```

But I guess you found out.

PR #3658, opened 2024-09-06
Fix - creation of missing reccord in table [Tenant Feature Key]

KM-JAD:
In case that related record in SQL table [dbo].[Tenant Feature Key] is missing, it is created automaticaly and then updated by this script as usual.


freddydk:
And you have tested that this actually works?

freddydk (approved):

PR #3670, opened 2024-09-18
Nuget fixes

freddydk:
NuGet feeds didn't work when using Containers / FilesOnly Containers

### DIFF
--- a/HelperFunctions.ps1
+++ b/HelperFunctions.ps1
@@ -1354,3 +1354,44 @@ function Write-GroupEnd {
         $bcContainerHelperConfig.IsAzureDevOps { Write-Host "##[endgroup]"; break }
     }
 }
+
+function CopySymbolsFromContainer {
+    Param(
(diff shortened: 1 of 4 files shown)


mazhelez (approved):

HelperFunctions.ps1
```diff
@@ -1354,3 +1354,44 @@ function Write-GroupEnd {
...
+        }
+        $paths | ForEach-Object {
+            $appFiles = $_.Split(',')
+            $appFile = ""
+            if (Test-Path -Path $appFiles[0]) {
+                $appFile = $appFiles[0]
+            }
+            elseif (Test-Path -path $appFiles[1]) {
+                $appFile = $appFiles[1]
+            }
```
jonasblunck, line 1388: Can be simplified into something like this:

$appFile = $_ -split "," | Where-Object { Test-Path ($_) } | Select -First 1

PR #3696, opened 2024-10-02
Enhancement of Misc/Set-BcContainerFeatureKeys.ps1 to be able to Enable Feature key for particular company.

KM-JAD:
There is added option to use new parameter "EnableInCompany" which triggers activation/deactivation of Feature Key for given company name. This is usefull in case you would like to activate Feature e.g. before autotests. Without this setup Feature itself doesn't work in any company.

**Important:** 
There is no activated any data upgrade. This must be initiated manually as usual (if it is really needed).


KM-JAD:

Misc/Set-BcContainerFeatureKeys.ps1
```diff
@@ -68,20 +82,37 @@ try {
                         $enabled = -1
                         Write-Host "WARNING: Unknown value ($enabledStr) for feature key $featureKey"
                     }
-                    if ($enabled -ne -1) {
+                    
+                    # Test if feature which has to be updated is available in table "Feature Data Update Status$63ca2fa4-4f03-4f2b-a480-172fef340d3f"
+                    $FeatureExistsInDestination = Invoke-Sqlcmd -Database $databaseName -Query $("SELECT COUNT(*) FROM [dbo].[Feature Data Update Status"+'$'+"63ca2fa4-4f03-4f2b-a480-172fef340d3f] where [Feature Key] = '$featureKey'")
+
+                    if(($FeatureExistsInDestination[0].ToString()) -eq "0")
+                    {
```
freddydk, line 90: The style in containerhelper all over is to place the { on the same line as the if statement - I know it is religion, but that is how it is all over and easier to read for me.

freddydk (changes requested):

freddydk:
Please also add a line to the upper section of the release notes

Misc/Set-BcContainerFeatureKeys.ps1
KM-JAD: sure...all adjusted

Misc/Set-BcContainerFeatureKeys.ps1
```diff
@@ -68,20 +82,37 @@ try {
...
                             $result = Invoke-Sqlcmd -Database $databaseName -Query "UPDATE [dbo].[Tenant Feature Key] set Enabled = $enabled where ID = '$featureKey';Select @@ROWCOUNT"
-                            if ($result[0] -eq "1") {
+                            
+                            # Update record in table "Feature Data Update Status$63ca2fa4-4f03-4f2b-a480-172fef340d3f" if it is requested for particular company
+                            $result2 = ''
+                            if (![String]::IsNullOrEmpty($EnableInCompany))
+                            {
+                                $result2 = Invoke-Sqlcmd -Database $databaseName -Query $("UPDATE [dbo].[Feature Data Update Status"+'$'+"63ca2fa4-4f03-4f2b-a480-172fef340d3f] set [Feature Status] = $enabled where [Feature Key] = '$featureKey' AND [Company Name] = '$EnableInCompany';Select @@ROWCOUNT")
+                            }
+                            if (($result[0] -eq "1") -and ((($result2[0] -eq "1") -and ![String]::IsNullOrEmpty($EnableInCompany)) -or ([String]::IsNullOrEmpty($EnableInCompany)))) {
```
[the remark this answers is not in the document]
KM-JAD: OK, code rebuilt and I think it is much easier now to read...

freddydk:
I added another write-host - but I think there is still something wrong here...

With the change in this PR - we only ever change anything if the feature exists in the table Feature Data Update Status`$63ca2fa4-4f03-4f2b-a480-172fef340d3f.
In my database, that table contains:

[bcserver]: PS C:\Run> Invoke-Sqlcmd -Database 'default' -Query "SELECT ""Feature Key"" FROM [dbo].[Feature Data Update Status`$63ca2fa4-4f03-4f2b-a480-172fef340d3f]"

Feature Key
-----------
SalesPrices
ExtensibleInvoicePostingEngine
GLCurrencyRevaluation
SalesPrices

None of these are in the tenant Feature key table.
Meaning that you would never insert anything in the Tenant Feature Key or like.

I also don't understand that the record should be in both databases,

Shouldn't it be like: If enabledInCompany - then we insert and/or modify one table - if not - then we insert and/or modify the other table.

KM-JAD:
1. Regarding new Write-Host change, OK, why not...I wanted to use your original logic of messages to join both tables changes into one message. Frankly speaking, I do not undertstand why you have changed "else" with new "if" on lines 115-117, but it seems that logic is still the same :)

2. Regarding missing records in table "Feature Data Update Status", well, the problem with Feature Management is a bit more complex.... There is an initial inconsistency in both tables "Tenant Feature Key" and "Feature Data Update Status" in original Cronus company provided within initial CRONUS database provided by MS.
In standard way, all respective lines in table [Feature Data Update Status`$63ca2fa4-4f03-4f2b-a480-172fef340d3f] are created by each particular feature CU based on published event in 2610-"Feature Management". Such approach is quite universal through BC versions, which is fully OK for normal database usage, but not good for autotests in case you would like to run autotests within container built on last version of artifacts. 
Therefore we added to our auto-test process two additional steps (out of Container helper world)
a) publishing page 2610 Feature Management as OData service (during extension for auto-tests installation process)
b) calling of this published OData service to generate all lines within table "Feature Data Update Status" (powershell, response is thrown)
In general, I didn't want to full rebuild your code....just wanted to make it working as initially intended.... :)

freddydk:
But... - with your change - the only values that can be used with this function are these:

SalesPrices
ExtensibleInvoicePostingEngine
GLCurrencyRevaluation
SalesPrices

as that is the content of the data update table. I cannot call this with any other key - I don't assume that that is the way it should be?

I would like to understand how this function is supposed to work - after all - I am the one needing to support this going forward.

Currently - the creation of a new record in Tenant Feature Key also doesn't work - as the SQL Query contains [CRONUS] database name as well - and that fails always - so the [CRONUS] part needs to go away I guess.

I will try to describe how I think this function should work (after fixing the CRONUS item above).

1. If called without a Companyname - it should create or update the feature key in [Tenant Feature Key] and set [Enabled] to 0|1 based on enabled.
2. If called with a Companyname it should do (1) and also create or update the feature key in [Feature Data Update Status`$63ca2fa4-4f03-4f2b-a480-172fef340d3f] and set [Feature Status] to 0|1 based on enabled.

Is that the intended functionality?

KM-JAD:
> as that is the content of the data update table. I cannot call this with any other key - I don't assume that that is the way it should be?

Yes, those four records are in database by default for company Cronus international, but you can create them also using configuration packages (instead of auto creation during opennig of page 2610)


> Currently - the creation of a new record in Tenant Feature Key also doesn't work - as the SQL Query contains [CRONUS] database name as well - and that fails always - so the [CRONUS] part needs to go away I guess.

well, yes, [CRONUS] is there the same way, as there is Business Central instance "BC' :) . Works now perfectly for containers, but I think it will work w/o that as well.


> I will try to describe how I think this function should work (after fixing the CRONUS item above).
> 
> 1. If called without a Companyname - it should create or update the feature key in [Tenant Feature Key] and set [Enabled] to 0|1 based on enabled.
> 2. If called with a Companyname it should do (1) and also create or update the feature key in [Feature Data Update Status`$63ca2fa4-4f03-4f2b-a480-172fef340d3f] and set [Feature Status] to 0|1 based on enabled.
> 
> Is that the intended functionality?

yes, it is :)

Please note that standard BC database downloaded from MS repository has a bit inconsistent data records in both our tables.

freddydk:
> well, yes, [CRONUS] is there the same way, as there is Business Central instance "BC' :) . Works now perfectly for containers, but I think it will work w/o that as well.

You specify -databaseName $databaseName and [CRONUS] - that doesn't work.
CRONUS is the databasename in a single tenant container - but the $databaseName is the name of a tenant.

KM-JAD:
> You specify -databaseName $databaseName and [CRONUS] - that doesn't work.
> CRONUS is the databasename in a single tenant container - but the $databaseName is the name of a tenant.

you are correct -> already fixed
Thanks for code revision :)

KM-JAD:
Hi Freddy, is there anything missing from my side?

Misc/Set-BcContainerFeatureKeys.ps1
```diff
@@ -68,23 +80,39 @@ try {
...
+                            #Create new record in table "Tenant Feature Key" in case it is missing
                             $SQLRecord = Invoke-Sqlcmd -Database $databaseName -Query "SELECT * FROM [dbo].[Tenant Feature Key] where ID = '$featureKey'"
-                            if ([String]::IsNullOrEmpty($SQLRecord))
-                            {
-                                Write-host "Creating record for feature ID '$featureKey'"
+                            if ([String]::IsNullOrEmpty($SQLRecord)) {
+                                Write-host "Creating record for feature $featureKey"
                                 $SQLcolumns = "ID, Enabled"
                                 $SQLvalues = "'$featureKey',0"
-                                Invoke-Sqlcmd -Database $databaseName -Query "INSERT INTO [CRONUS].[dbo].[Tenant Feature Key] ($SQLcolumns) VALUES ($SQLvalues)" -Verbose
```
MODUSCarstenScholling, line 80: Not sure about the whole thing, but this missing change currently breaks a pipeline when trying to set/add feature keys, because of the database name.

PR #3699, opened 2024-10-03
Fix bug in Download-BcNuGetPackageToFolder

adamLiveware:
Error occurs when using -downloadDependencies with value 'own' because of an invalid reference

This fixes that by correcting the reference, allowing to download dependencies where the author matches


freddydk (approved):

freddydk:
Thanks

PR #3700, opened 2024-10-04
Version number fixes

freddydk:
Three version number fixes:

- Bump version
- As minimum, always use the generic tag version which was available when shipping BcContainerHelper
- Add awareness of Windows 11 24H2


f4n0:
@freddydk We have an ETA about supporting the 24H2 with process isolation?

freddydk:
> @freddydk We have an ETA about supporting the 24H2 with process isolation?

We can try - if it works, it works. It is really out of my hands.

aholstrup1 (approved):

PR #3710, opened 2024-10-09
Copy Business Foundation Test Libraries app to symbols folder in New-BcCompilerFolder

skeck:
I probably found a missing app handling in `New-BcCompilerFolder`. When populating the cache folder, "Business Foundation Test Libraries" is not being copied to the symbols folder. Therefore, apps that have a dependency on it will not compile due to missing symbols.

Example:
```
$artifactURL = "https://bcartifacts-exdbf9fwegejdqak.b02.azurefd.net/sandbox/25.0.23364.25278/de"
$compilerFolder = New-BcCompilerFolder -artifactUrl $artifactURL -includeAL
$symbolsFolder = Join-Path $compilerFolder "symbols"

$filePattern = "Microsoft_Business Foundation Test Libraries*.app"
$fileCheck = Get-ChildItem -Path $symbolsFolder -Filter $filePattern
if ($fileCheck) {
    Write-Host "File exists in the folder: $symbolsFolder" # result with fix
} else {
    Write-Host "File does NOT exist in the folder: $symbolsFolder"  # result without the fix
}
```

### DIFF
--- a/CompilerFolderHandling/New-BcCompilerFolder.ps1
+++ b/CompilerFolderHandling/New-BcCompilerFolder.ps1
@@ -146,7 +146,7 @@ try {
             if ($appAppsPath) {
                 $appApps = @(Get-ChildItem -Path $appAppsPath -Filter '*.app' -Recurse)
             }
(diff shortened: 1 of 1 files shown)


freddydk:
It looks like you are right - and we would also need to add `Microsoft_AI Test Toolkit` to that as well

freddydk (approved):

freddydk:
Thanks

PR #3739, opened 2024-10-26
Increase Get-BcArtifactUrl Performance

freddydk:
When looking for the latest version of the artifacts for a specific country, we need to fetch all artifacts in the blob storage in order to figure out what is latest. This fix uses an approximation of latest version number (actually the one shipping in a month) and trying to download that (or 2 prior versions)

Additional improvements we are looking at for future optimizations:

- Use OCI artifacts for artifacts storage instead of traditional blob storage
- Restructure artifacts in more layers to allow for partial download of what you need in various scenarios
- Create index if artifacts to allow for querying artifacts without reading tags/annotations


BC.HelperFunctions.ps1
```diff
@@ -118,6 +118,7 @@ function Get-ContainerHelperConfig {
             "IsGitHubActions" = ($env:GITHUB_ACTIONS -eq "true")
             "IsAzureDevOps" = ($env:TF_BUILD -eq "true")
             "IsGitLab" = ($env:GITLAB_CI -eq "true")
+            "useApproximateVersion" = $true
```
aholstrup1, line 121: Should be an opt-in and not an opt-out IMO. At least to begin with. If we find a bug in it the next few weeks it won't impact everyone

Artifacts/Get-BCArtifactUrl.ps1
```diff
@@ -141,172 +141,206 @@ try {
...
-        elseif ($select -eq 'Closest') {
-            if (!($version)) {
-                throw "You must specify a version number when you want to get the closest artifact Url"
+
+        if ($type -eq 'sandbox' -and $storageAccount -eq 'bcartifacts.blob.core.windows.net' -and $select -eq 'latest' -and $version -eq '' -and $bcContainerHelperConfig.useApproximateVersion) {
+            # Temp fix / hack for Get-BcArtifact performance
+            # If Microsoft changes versioning schema, this needs to change (or useApproximateVersion should be set to false)
+            $now = ([DateTime]::Now).AddDays(15)
+            $approximateMajor = 23+2*($now.Year-2024)+($now.Month -ge 4)+($now.Month -ge 10)
+            $approximateMinor = ($now.Month - 4)%6
```
aholstrup1, line 150: Won't this become negative in January? What happens when we look for 25.-3?

freddydk: you are right

aholstrup1 (approved):

mazhelez (approved):

Artifacts/Get-BCArtifactUrl.ps1
```diff
@@ -141,172 +141,206 @@ try {
...
-            }
-        }
-        elseif ($select -eq 'Closest') {
-            if (!($version)) {
-                throw "You must specify a version number when you want to get the closest artifact Url"
+
+        if ($type -eq 'sandbox' -and $storageAccount -eq 'bcartifacts.blob.core.windows.net' -and $select -eq 'latest' -and $version -eq '' -and $bcContainerHelperConfig.useApproximateVersion) {
+            # Temp fix / hack for Get-BcArtifact performance
+            # If Microsoft changes versioning schema, this needs to change (or useApproximateVersion should be set to false)
+            $now = ([DateTime]::Now).AddDays(15)
```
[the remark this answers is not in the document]
freddydk: the 15 days ahead is because we typically ship on the 10th - 12th - so it is a better starting approximation

ReleaseNotes.txt
```diff
@@ -1,3 +1,6 @@
+6.0.28
```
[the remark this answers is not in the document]
freddydk: Yes, will update that in 2 sec.

Issue #3590, opened 2024-07-05
Docs on how to handle NuGet Runtime Packages 


lktraser:
**Describe the issue**
Hi this is not an issue in the sense of it's not working but missing documentation or pointers to such documentation.

I'm trying to adopt the NuGet scripts for our CI/CD and had a really hard time figuring out how to run the scripts correctly for Runtimes to work as intented. I'll include my script below because I need someone to tell me if I did this right or if I'm missing the obvious easy path I should have taken. I worked based on the slides from the BC TechDays but even with that information it still took me hours to get it right. I don't think such a fundamental thing should be this hard so we make the barrier of entry lower for better adoption rates.

My proposal would be to A) Write a post for Freddy's Blog on how to Push and Pull Runtimes with NuGet and B) include examples for runtimes in the documentation headers of the Push and Pull Scripts. Some general guidelines on how to setup Feeds, (prelease, release, runtimes, internal, external would also be appreciated)

I would really like to see large scale adoption of this technologie but handling runtimes is still a big part of that and needs to be easy. Ideally there should be a cmdlet/parameter that just does it correctly out of the box. 

I've used the slide below to get this far. Is it correct that runtimes can only be pulled with 'Download-BcNuGetPackageToFolder' because 'Get-BcNuGetPackage' doesn't explore the dependencies and therefore will never find the actual apps?

**Scripts used to create container and cause the issue**
Pushing Runtimes:
```
foreach ($app in $apps.FullName) {
    $appJson = Get-AppJsonFromAppFile -appFile $app

    $packageName = Get-BcNuGetPackageId -publisher $appJson.publisher -name $appJson.name -id $appJson.id -version $appJson.version -tag 'runtime'

    $feed, $packageId, $packageVersion = Find-BcNuGetPackage `
        -nuGetServerUrl $nugetserverurl `
        -nuGetToken $nugetToken `
        -packageName $packageName `
        -version $appJson.version `
        -select Exact

    if ($packageVersion -eq $appJson.version) {
        continue
    }

    $NugetPackage = New-BcNuGetPackage `
        -appfile $app `
        -isIndirectPackage `
        -packageId "{publisher}.{name}.runtime.{id}" `
        -dependencyIdTemplate "{publisher}.{name}.runtime.{id}" #-prereleaseTag 'runtime'

    Push-BcNuGetPackage -nuGetServerUrl $nugetserverurl -nuGetToken $nugetToken -bcNuGetPackage $NugetPackage

[...]

PR #3759, opened 2024-11-13
normalize nuget version on FindPackageVersion

lktraser:
#3590

fixing FindPackageVersion for runtime packages with versions with trailing 0  in build or revision.

### DIFF
--- a/NuGet/NuGetFeedClass.ps1
+++ b/NuGet/NuGetFeedClass.ps1
@@ -277,7 +277,7 @@ class NuGetFeed {
             if ($excludeVersions -contains $version) {
                 continue
             }
-            if (($select -eq 'Exact' -and [NuGetFeed]::NormalizeVersionStr($nuGetVersionRange) -eq [NuGetFeed]::NormalizeVersionStr($version)) -or ($select -ne 'Exact' -and [NuGetFeed]::IsVersionIncludedInRange($version, $nuGetVersionRange))) {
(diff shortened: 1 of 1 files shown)


NuGet/NuGetFeedClass.ps1
```diff
@@ -277,7 +277,7 @@ class NuGetFeed {
             if ($excludeVersions -contains $version) {
                 continue
             }
-            if (($select -eq 'Exact' -and [NuGetFeed]::NormalizeVersionStr($nuGetVersionRange) -eq [NuGetFeed]::NormalizeVersionStr($version)) -or ($select -ne 'Exact' -and [NuGetFeed]::IsVersionIncludedInRange($version, $nuGetVersionRange))) {
+            if (($select -eq 'Exact' -and [NuGetFeed]::NormalizeVersionStr($nuGetVersionRange) -eq [NuGetFeed]::NormalizeVersionStr($version)) -or ($select -ne 'Exact' -and [NuGetFeed]::IsVersionIncludedInRange([NuGetFeed]::NormalizeVersionStr($version), [NuGetFeed]::NormalizeVersionStr($nuGetVersionRange)))) {
```
freddydk, line 280: I don't think you can use the NormalizeVersionStr on the version range, which can be "[24.4.0.0)" - or?

freddydk (changes requested):

NuGet/NuGetFeedClass.ps1
freddydk: Looking in the IsVersionIncludedInRange function - that one already uses NormalizeVersionStr on the individual version numbers found.

lktraser: I was wondering about that to but the first part of the line also used Normalization on the Range so I didn't look too hard at it.

lktraser: Absolutely correct, I reran my tests to see if I just phantom fixed the issue with this and that seems to be the case. 
I'll abandon this.

PR #3760, opened 2024-11-13
[BcNuGet] Reduce memory consumption, improve performance, prevent memory leaks

lippertmarkus:
We found that `Compress-Archive` and especially storing the `octet-stream` inside a string instead of using a Stream causes a very high memory consumption and a bad performance.

We're invoking the BcNuGet Cmdlets from our own C# API and therefore were able to profile the memory and performance. The behaviour however is similar when invoking the Cmdlets directly (PowerShell Core 7.4.6).

**Memory (Note: around 340 MB is the base load our API brings with it)**

Before: Memory spikes when uploading big apps, around 1 GB memory consumption from BcNuGet Cmdlets
image

After: 
image

**Performance**

Before:
image

After:
image

### DIFF
--- a/NuGet/New-BcNuGetPackage.ps1
+++ b/NuGet/New-BcNuGetPackage.ps1
@@ -286,7 +286,11 @@ Function New-BcNuGetPackage {
         if (Test-Path $nuPkgFile -PathType Leaf) {
             Remove-Item $nupkgFile -Force
         }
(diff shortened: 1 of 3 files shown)


freddydk (changes requested):
Looks good - could you add a line or two to releasenotes.txt explaining what has changed (just high level as everything else)

lippertmarkus:
I added releasenotes

PR #3763, opened 2024-11-14
Give Option to Choose SQL PowerShell Module When Restoring From BacPac

### LINKED ISSUE #3762: Need to provide parameter for which SQL PowerShell module to use
**Describe the issue**
There can be issues when both SQLPS and the SQLServer PowerShell modules are both installed. Restore-BcDatabaseFromArtifacts will fail. Additionally I believe SQLPS is not the recommended module at this point.

**Scripts used to create container and cause the issue**
Following the scripts here: https://community.dynamics.com/blogs/post/?postid=cb3372ed-161f-4549-b935-7b4028e0cd72

Proposed PR coming shortly.

MattTraxinger:
Fixes #3762 

This is a change to how Restore-BcDatabaseFromArtifacts chooses which PowerShell module to use. I have added an optional parameter, SqlModuleToUse, with two possible values, sqlps and sqlserver, and a default value of sqlps to preserve existing functionality.

I've also added the same parameter to New-BcContainer.

### DIFF
--- a/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
+++ b/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
@@ -23,7 +23,7 @@
  .Parameter sqlTimeout
   SQL Timeout for database restore operations
  .Parameter useSqlServerModule
-  Switch, forces the use of the sqlserver module instead of the sqlps module
+  Switch, forces the use of the sqlserver module instead of the sqlps module. Default is to use the sqlps module. The default can be changed in the bcContainerHelperConfig file by setting "useSqlServerModule" = $false.
 #>
 function Restore-BcDatabaseFromArtifacts {
     Param(
@@ -42,7 +42,7 @@ function Restore-BcDatabaseFromArtifacts {
         [switch] $multitenant,
         [switch] $async,
         [int] $sqlTimeout = -1,
-        [switch] $useSqlServerModule
+        [switch] $useSqlServerModule = $bcContainerHelperConfig.useSqlServerModule
     )
 
 $telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
@@ -101,6 +101,7 @@ try {
             if ($multitenant) {
                 $dbName = "$($databasePrefix)tenant"
             }
+            $sqlpsModule = $null
             if(-not $useSqlServerModule) {
                 Import-Module sqlps -ErrorAction SilentlyContinue
                 $sqlpsModule = get-module sqlps
@@ -227,7 +228,7 @@ try {
             throw
         }
     
-    } -ArgumentList $containerHelperPath, $artifactUrl, $databaseServer, $databaseInstance, $databasePrefix, $databaseName, $multitenant, $successFileName, $bakFile, $sqlTimeout, $useSqlServerModule
+    } -ArgumentList $containerHelperPath, $artifactUrl, $databaseServer, $databaseInstance, $databasePrefix, $databaseName, $multitenant, $successFileName, $bakFile, $sqlTimeout, $useSqlServerModule.IsPresent
 
     if (!$async) {
         While ($job.State -eq "Running") {
--- a/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
+++ b/Bacpac/Restore-BcDatabaseFromArtifacts.ps1
@@ -22,8 +22,8 @@
   Include this parameter if you want to restore the database asynchronous. A file called <databasePrefix>databasescreated.txt will be created in the containerhelper folder when done
  .Parameter sqlTimeout
   SQL Timeout for database restore operations
- .Parameter sqlModuleToUse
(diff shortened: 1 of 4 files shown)


MattTraxinger:
@microsoft-github-policy-service agree [company="ArcherPoint"]

MattTraxinger:
@microsoft-github-policy-service agree company="ArcherPoint"

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
```diff
@@ -98,14 +101,17 @@ try {
...
-                import-module SqlServer
-                $SqlModule = get-module SqlServer
-                if (-not $SqlModule) {
-                    throw "You need to have a local installation of SQL or you need the SqlServer PowerShell module installed"
-                }
+            if($sqlModuleToUse -eq "sqlps") {
+                Import-Module sqlps
+                $SqlModule = Get-Module sqlps
+                Write-Host "Using sqlps module."
+            } elseif ($sqlModuleToUse -eq "sqlserver") {
```
freddydk, line 108: Current functionality will try sqlps and if that is not found - then it will try sqlserver (which will work in PS7 where SQLPS is not available). The new functionality will break this behavior and make the function always fail on PS7

A better (and non-breaking) fix would be to replace the sqlModuleToUse with a switch parameter called UseSqlServerModule - and then not try sqlps - which I guess is what you want?

freddydk: And then maybe create a setting called UseSqlServerModule - which you use as default value for the parameter - meaning that people can set this setting on the machine to always use SqlServer

freddydk (changes requested):

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
MattTraxinger: Will do.

Yes. To be more specific about my problem, sqlps always fails for for me, while sqlserver always succeeds. This has held true for me on different machines and different versions of PowerShell for several months. If I uninstall sqlps, it just comes back when I run the existing BcContainerHelper scripts and never even tries to use sqlserver. At this point, I just want to make the choice for myself.

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
```diff
@@ -222,7 +227,7 @@ try {
             throw
         }
     
-    } -ArgumentList $containerHelperPath, $artifactUrl, $databaseServer, $databaseInstance, $databasePrefix, $databaseName, $multitenant, $successFileName, $bakFile, $sqlTimeout
+    } -ArgumentList $containerHelperPath, $artifactUrl, $databaseServer, $databaseInstance, $databasePrefix, $databaseName, $multitenant, $successFileName, $bakFile, $sqlTimeout, $useSqlServerModule
```
freddydk, line 230: use $useSqlServerModule.IsPresent here.

Bacpac/Restore-BcDatabaseFromArtifacts.ps1
```diff
@@ -98,8 +101,10 @@ try {
             if ($multitenant) {
                 $dbName = "$($databasePrefix)tenant"
             }
-            Import-Module sqlps -ErrorAction SilentlyContinue
-            $sqlpsModule = get-module sqlps
+            if(-not $useSqlServerModule) {
+                Import-Module sqlps -ErrorAction SilentlyContinue
+                $sqlpsModule = get-module sqlps
+            }
```
freddydk, line 107: sqlpsmodule is uninitialized in 108 I think?

ContainerHandling/New-NavContainer.ps1
```diff
@@ -526,7 +529,7 @@ try {
                 $multitenant = $bcContainerHelperConfig.sandboxContainersAreMultitenantByDefault
             }
             Remove-BcDatabase -databaseServer $databaseServer -databaseInstance $databaseInstance -databaseName "$($databasePrefix)%"
-            Restore-BcDatabaseFromArtifacts -artifactUrl $artifactUrl -databaseServer $databaseServer -databaseInstance $databaseInstance -databasePrefix $databasePrefix -databaseName $databaseName -multitenant:$multitenant -bakFile $bakFile -async
+            Restore-BcDatabaseFromArtifacts -artifactUrl $artifactUrl -databaseServer $databaseServer -databaseInstance $databaseInstance -databasePrefix $databasePrefix -databaseName $databaseName -multitenant:$multitenant -bakFile $bakFile -useSqlServerModule $useSqlServerModule -async
```
freddydk, line 532: a switch needs a colon - i.e. -useSqlServerModule:$useSqlServerModule.IsPresent

freddydk (changes requested):
A few comments
Add a setting in Bc.HelperFunctions.ps1 (like useApproximateVersion) and use that as default for the two functions.
plus add a line to release notes.

freddydk:
sorry if my review was too early:-)

freddydk (approved):

PR #3765, opened 2024-11-20
EarliestMatching search for Download-BcNuGetPackageToFolder

lktraser:
For runtime creation / OnPrem deployments it's usually needed to retrieve runtime packages with the earliest matching version. This approach feels almost too easy so please point out if I missed something but it seemed to resolve the way I want it to.

#3590

### DIFF
--- a/NuGet/Download-BcNuGetPackageToFolder.ps1
+++ b/NuGet/Download-BcNuGetPackageToFolder.ps1
@@ -254,7 +254,7 @@ try {
                         throw $dependenciesErr
                     }
                     else {
-                            # If we are looking for the earliest/latest matching version, then we can try to find another version
+                        # If we are looking for the earliest/latest matching version, then we can try to find another version
                         Write-Host "WARNING: $dependenciesErr"
                         break
                     }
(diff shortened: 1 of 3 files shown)


freddydk:
Generally looks good - a few things missing though:

- [ ] In the parameter description for select, EarliestMatching should be added
- [ ] There are a few comments mentioning "latest matching", which should be latest or earliest matching
- [ ] Please add a line to the release notes under 6.0.30
- [ ] Please add the same parameter description and value to the validateSet of Publish-BcNuGetPackageToContainer

freddydk (changes requested):
Generally looks good - a few things missing though:

- [x] In the parameter description for select, EarliestMatching should be added
- [x] There are a few comments mentioning "latest matching", which should be latest or earliest matching
- [x] Please add a line to the release notes under 6.0.30
- [ ] Please add the same parameter description and value to the validateSet of Publish-BcNuGetPackageToContainer

lktraser:
Added the requested changes. Hope I found all comments refering to match search.

freddydk (approved):

PR #3784, opened 2024-12-07
Fix for missing Label for platform on docker container

slaven:
A minor fix that should be implemented in case of missing docker container Label `platform`.

`Get-BcContainerPlatformVersion` can return an empty string, which results in as a soft error, hard to track because it runs inside of container.

The error:
`Cannot convert value "" to type "System.Version". Error: "Version string portion was too short or too long. (Parameter 'input')"`

I tried to be consistent how you handle soft errors for `Get-BcContainerPlatformVersion`. I used `Get-BcContainerNavVersion` as a fallback and additionally added a warning to point the user in the right direction to fix it.


freddydk:
In what scenario would you have a container without the platform label?

PR #3807, opened 2025-01-17
Extend condition to check for name and publisher

AlexCeg:
We have the problem, that we have a app which is named "System" as well. The problem is that we get the following error when we use the funciton `Compile-AppWithBcCompilerFolder`: 
image

Therefore we need to also check the publisher of the dependency to avoid the error and wrong behaviour in the other places.


freddydk (approved):

freddydk:
Thanks

AlexCeg:
@freddydk thank you for the quick and easy processing.

freddydk:
Thanks for the PR:-)

PR #3808, opened 2025-01-17
Support newer artifacts with lower case folder names

freddydk:
Fixes https://github.com/microsoft/AL-Go/issues/1400


ReleaseNotes.txt
```diff
@@ -1,4 +1,5 @@
 6.0.31
+New-BcCompilerFolder with base artifacts doesn't loate apps in version 25.0 due to a rename of the applications folder (was Applications)
```
mazhelez, line 2: Typo

```suggestion
New-BcCompilerFolder with base artifacts doesn't locate apps in version 25.0 due to a rename of the applications folder (was Applications)
```

mazhelez (approved):

PR #3832, opened 2025-02-12
CompilerFolder doesn't transfer installed Apps to NuGet resolution

freddydk:
This PR is the BcContainerHelper part of making sure that the installed apps, the platform and the country are transferred to NuGet resolution when locating apps for downloading - making sure that you don't download a package with dependencies to BC which is newer than what you are building for.

It is a requirement for https://github.com/microsoft/AL-Go/pull/1450

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -508,7 +508,7 @@ function GetInstalledApps {
     Write-GroupStart -Message "Installed Apps"
     $installedApps | ForEach-Object {
         Write-Host "- $($_.AppId):$($_.Name)"
-        return @{ "AppId" = "$($_.AppId)"; "Name" = "$($_.Name)"; "Publisher" = "$($_.Publisher)"; "Version" = "$($_.Version)" }
+        return @{ "Id" = "$($_.AppId)"; "Name" = "$($_.Name)"; "Publisher" = "$($_.Publisher)"; "Version" = "$($_.Version)" }
     }
     Write-GroupEnd
 }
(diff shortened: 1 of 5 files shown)


AppHandling/Run-AlPipeline.ps1
```diff
@@ -506,7 +508,7 @@ function GetInstalledAppIds {
     Write-GroupStart -Message "Installed Apps"
     $installedApps | ForEach-Object {
         Write-Host "- $($_.AppId):$($_.Name)"
-        "$($_.AppId)"
+        return @{ "Id" = "$($_.AppId)"; "Name" = "$($_.Name)"; "Publisher" = "$($_.Publisher)"; "Version" = "$($_.Version)" }
```
jonasblunck, line 511: The code in GetInstalledAppIds adds the field "AppId", but this code here uses "Id". Align?

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1655,6 +1659,8 @@ Measure-Command {
     $Parameters = @{
         "missingDependencies" = @($unknownTestAppDependencies | Where-Object { $missingTestAppDependencies -contains "$_".Split(':')[0] })
         "appSymbolsFolder" = $appSymbolsFolder
+        "installedApps" = $installedApps
+        "installedCountry" = $artifactUrl.Substring($artifactUrl.LastIndexOf('/')+1)
```
jonasblunck, line 1663: I see some duplication of code in '$artifactUrl.Substring($artifactUrl.LastIndexOf('/')+1)'. Move out to separate function?

AppHandling/Run-AlPipeline.ps1
freddydk: There are 3 different ways to get the installed apps - and they return different objects - this function will return this in the way needed by the rest of containerhelper and AL-Go, a hashtable with id, name, version and publisher.

AppHandling/Run-AlPipeline.ps1
freddydk: You will find a lot of things like this in Run-AlPipeline. In the end, the function is being deprecated, I don't want to start cleaning up things like this now.

aholstrup1 (approved):

Issue #3831, opened 2025-02-12
New-BcCompilerFolder parameter vsixFile does not consider a local path


MODUSCarstenScholling:
**Describe the issue**
The New-BcCompilerFolder parameter vsixFile does not consider a local path. This probably applies to additional areas.

[See here](https://github.com/microsoft/navcontainerhelper/blob/b2344b2d8d23cc3401bff59b70552914f21b470f/CompilerFolderHandling/New-BcCompilerFolder.ps1#L203)

Would you allow a PR changing this (I will check other areas, too)?

- We have 20-30% pipeline fails due to timeouts downloading the AL Language vsix when running Run-AlPipeline.
- Long running downloads should be avoided

**Full output of scripts**

Same with BCCH 6.0.33.

Example for timeout:
```
2025-02-12T05:35:50.3913181Z Creating CompilerFolder
2025-02-12T05:35:56.7241784Z Using https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/14.2.1249978/1733905314434/Microsoft.VisualStudio.Services.VSIXPackage
2025-02-12T05:35:56.7255695Z Downloading C:\Users\ItsMeYourBrother\AppData\Local\Temp\alc.A28-s33865-de25-0-23364-29957.zip
2025-02-12T05:37:57.0047649Z Error downloading..., retrying in 6 seconds...
2025-02-12T05:40:03.1550978Z ##[group]Removing container
2025-02-12T05:40:03.3349538Z Removing Desktop shortcuts
2025-02-12T05:40:03.3444034Z ##[endgroup]
2025-02-12T05:40:04.0938119Z Run-AlPipeline Telemetry Correlation Id: b4ed7228-d0c9-4292-a184-456a358b87ce
2025-02-12T05:40:04.1661329Z Timeout downloading file at Download-File, C:\Users\ItsMeYourBrother\Documents\PowerShell\Modules\BcContainerHelper\6.0.29\Common\Download-File.ps1: line 96
2025-02-12T05:40:04.1676701Z at New-BcCompilerFolder, C:\Users\ItsMeYourBrother\Documents\PowerShell\Modules\BcContainerHelper\6.0.29\CompilerFolderHandling\New-BcCompilerFolder.ps1: line 195
[...]

PR #3834, opened 2025-02-17
Reworked AL Language extension handling, compiler and cache folder

MODUSCarstenScholling:
Reworked AL Language extension handling (issue #3831 ).
Download to folder $bcContainerHelperConfig.hostHelperFolder/alLanguageExtension. Sym-linked into compiler and cache folder.

If a vsixFile is specified for processes, it is downloaded and extracted (if requested) to the alLanguageExtension folder first. 
If no vsixFile is specified, the default behavior is used, and the compiler folder (and cache folder) is populated with the artifacts AL Language extension.


freddydk:
Will have a look today and/or tomorrow.
QQ - have you tested the symbolic link functionality on Linux?

freddydk:
> colleague to test symlinks on his MacOS and I will try it later today. The docs say: works. But it is the PS7 docs, so there is no guarantee ;-)

I will switch a local AL-Go version to run on this ContainerHelper and test out builds on Linux as well.
Note that I am working on re-writing the entire artifacts download mechanism, so you shouldn't dive too deep into that area.
We are moving away from Storage Accounts and over to OCI artifacts - which will makes things much faster and more reliable.

ContainerHandling/Get-LatestAlLanguageExtensionUrl.ps1
```diff
@@ -11,21 +15,27 @@
...
-    return $url
-}
-catch {
-    TrackException -telemetryScope $telemetryScope -errorRecord $_
-    throw
-}
-finally {
-    TrackTrace -telemetryScope $telemetryScope
-}
+	$telemetryScope = InitTelemetryScope -name $MyInvocation.InvocationName -parameterValues $PSBoundParameters -includeParameters @()
```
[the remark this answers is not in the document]
MODUSCarstenScholling: That probably does indeed seem a bit odd...  I have not found another solution to properly identify the version of a real path, including the version as string. So the vsix file is passed to extract the version for lookup. This allows to identify a prerelease.

ContainerHandling/New-NavContainer.ps1
```diff
@@ -1271,24 +1272,7 @@ try {
...
-        elseif (Test-Path $vsixUrl -PathType Leaf) {
-            $vsixFile = "$containerFolder\$([System.IO.Path]::GetFileName($vsixUrl))"
-            Copy-Item -Path $vsixUrl -Destination $vsixFile
-        }
-        else {
-            $vsixFile = ""
-            throw "Unable to locate vsix file ($vsixUrl)"
-        }
-    }
+	$vsixFile = GetAlLanguageExtension -vsixFile $vsixFile;
```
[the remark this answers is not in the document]
MODUSCarstenScholling: if I remember correctly, because the file is copied to the container if specified.

MODUSCarstenScholling:
Latest commit [1b8008f]: I had this version running on 11 agents without issues until now. It seems that we need to difference between symbolic link and folder.

ContainerHandling/Get-LatestAlLanguageExtensionUrl.ps1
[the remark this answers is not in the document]
freddydk: The function is designed to return a url to the AL Language extension in the market place - either latest or preview.
it shouldn't download anything or have any extra parameter - is this because the function now downloads something???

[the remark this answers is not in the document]
freddydk: So, now this function returns a file:// url with the location of the vsix???
I think that is too intrusive - this function might be used for other purposes in other places and now does something completely different.
Things like this might give me support problems with many other partners.
I will schedule time later today to do a review and understand what the suggested changes are doing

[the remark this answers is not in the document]
MODUSCarstenScholling: Ok, it seems that I did not properly describe the change. Let me show you some examples instead. See, it is non-breaking

```PowerShell
PS> Get-LatestAlLanguageExtensionUrl # Returns the latest regular VSIX URL like before
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/14.2.1249978/1733905314434/Microsoft.VisualStudio.Services.VSIXPackage
PS> Get-LatestAlLanguageExtensionUrl -allowPrerelease # Returns the latest prerelease VSIX like before
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/15.0.1338262/1740411038495/Microsoft.VisualStudio.Services.VSIXPackage
PS> Get-LatestAlLanguageExtensionUrl -allowPrerelease -vsixFile 'previous' # NEW: Returns the previous prerelease VSIX
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/15.0.1290655/1737129403467/Microsoft.VisualStudio.Services.VSIXPackage
PS> Get-LatestAlLanguageExtensionUrl -vsixFile 'previous' # NEW: Returns the previous VSIX
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/14.1.1180850/1729100221227/Microsoft.VisualStudio.Services.VSIXPackage
PS> Get-LatestAlLanguageExtensionUrl -allowPrerelease -vsixFile 'C:\ProgramData\BcContainerHelper\alLanguageExtension\14.2.1249978' # Extracts the version, matches it with MP, allowPrerelease does not matter, returns URL
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/14.2.1249978/1733905314434/Microsoft.VisualStudio.Services.VSIXPackage
PS> Get-LatestAlLanguageExtensionUrl -vsixFile 'C:\ProgramData\BcContainerHelper\alLanguageExtension\14.2.1249978.vsix' # Extracts the version, matches it with MP, returns URL
https://ms-dynamics-smb.gallerycdn.vsassets.io/extensions/ms-dynamics-smb/al/14.2.1249978/1733905314434/Microsoft.VisualStudio.Services.VSIXPackage
```

MODUSCarstenScholling:
@freddydk Is there anything I can do to help you getting this in?

freddydk:
Sorry, the review was a little bigger than anticipated and I had a vacation to attend.
Back tomorrow and will have a look again.

CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -51,12 +51,12 @@ try {
     $version = [System.Version]($parts[4])
     $country = $parts[5]
 
-    $vsixFile = DetermineVsixFile -vsixFile $vsixFile
-
     if ($version -lt "16.0.0.0") {
         throw "Containerless compiling is not supported with versions before 16.0"
     }
     
+	$vsixFile = GetAlLanguageExtension -vsixFile $vsixFile -extract;
```
freddydk, line 58: If running on a GitHub hosted runner (i.e. no local cache), you will be downloading and extracting the AL Language extention with every compiler folder and basically disabling the GitHub cache.
The cacheFolder might already be populated with compiler, apps and everything when we get here.

Maybe you could move this to line 81 - and only do this if $compilerPath doesn't exist ?

CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -195,17 +195,23 @@ try {
...
         Write-Host "Using $vsixFile"
-        $tempZip = Join-Path ([System.IO.Path]::GetTempPath()) "alc.$containerName.zip"
-        Download-File -sourceUrl $vsixFile -destinationFile $tempZip
-        Expand-7zipArchive -Path $tempZip -DestinationPath $containerCompilerPath
+
+		DeletePathOrSymbolicLink -path $containerCompilerPath;
+		New-Item -Path $containerCompilerPath -ItemType SymbolicLink -Value $vsixFile | Out-Null;
+
+		if ($cacheFolder) {
+			DeletePathOrSymbolicLink -path $compilerPath;
```
freddydk, line 208: Will the GitHub cache cache files behind the symbolic link?
cacheFolder here is the folder, which is cached to the GitHub cache if the workflow is successful - this needs to include the compiler as well?
I wouldn't expect it to - so maybe if we are extracting to a cacheFolder - we need to copy?

HelperFunctions.ps1
```diff
@@ -1207,28 +1207,133 @@ function GetAppInfo {
...
+		[string] $vsixFile = 'latest',
+        [switch] $allowPrerelease
+    )
+
+	$listing = Invoke-WebRequest -Method POST -UseBasicParsing `
+		-Uri 'https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery?api-version=3.0-preview.1' `
+		-Body '{"filters":[{"criteria":[{"filterType":8,"value":"Microsoft.VisualStudio.Code"},{"filterType":12,"value":"4096"},{"filterType":7,"value":"ms-dynamics-smb.al"}],"pageNumber":1,"pageSize":50,"sortBy":0,"sortOrder":0}],"assetTypes":[],"flags":0x192}' `
+		-ContentType 'application/json' | ConvertFrom-Json;
+	
+	$preReleaseKey = 'Microsoft.VisualStudio.Code.PreRelease';
```
freddydk, line 1221: I hate semicolons at the end of each line...

ContainerHandling/Get-LatestAlLanguageExtensionUrl.ps1
```diff
@@ -3,6 +3,10 @@
   Get the URL of the latest AL Language Extension from VS Code Marketplace
  .Description
   Get the URL of the latest AL Language Extension from VS Code Marketplace
+ .Parameter vsixFile
```
freddydk, line 6: There is a new function, which downloads and extracts the latest Vsix from marketplace - this function should not - it should "just" return the URL as it used to.
I am simply unable to explain to people why the vsixFile parameter is added to this GetUrl function.

AppHandling/Run-AlPipeline.ps1
```diff
@@ -1066,7 +1067,7 @@ if ($installOnlyReferencedApps) {
     $installTestApps += @($installApps)
 }
 
-$vsixFile = DetermineVsixFile -vsixFile $vsixFile
+$vsixFile = GetAlLanguageExtension -vsixFile $vsixFile -extract;
```
freddydk, line 1070: Extracting the vsixFile here will disable the GitHub cache - on GitHub hosted runners, this will always download and extract the vsix file

AppHandling/Run-AlValidation.ps1
```diff
@@ -295,8 +296,7 @@ else {
     $RemoveBcContainer = { Param([Hashtable]$parameters) Remove-BcContainer @parameters }
 }
 
-$vsixFile = DetermineVsixFile -vsixFile $vsixFile
-
+$vsixFile = GetAlLanguageExtension -vsixFile $vsixFile -extract;
```
freddydk, line 299: Extracting the vsixFile here will disable the GitHub cache - on GitHub hosted runners, this will always download and extract the vsix file

freddydk (changes requested):
There is a problem with downloading and extracting the vsixFile early all the time - this shouldn't be done until we actually need the vsixfile.
We should carry around the name of the vsixFile we need until we actually need it and then either use it from the github cache (if already there), the local cache (which you introduce here), the container (if that's where it is taken from) or download it.

I hope this makes sense.

MODUSCarstenScholling:
@freddydk Thank you for this in-depth review. It seems that the gearings between BCCH and AL-Go are too close to change something like this. Based on your comments and a review of requirements for AL-Go, I have decided to implement this in our pipeline solution and leave BCContainerHelper as it is. 

There would be changes needed in AL-Go to make this work that I am not yet able to do. I am closing this PR.

Thank you again and sorry for the time you invested. On the other hand, I am a bit deeper in AL-Go now, 0.1% more compared to Tuesday :)

freddydk:
I don't think anything would need to be modified in AL-Go to make this caching work - it is merely a problem with populating the cache early or just-in-time.
In your PR here - you download and populate the cache up-front - instead of waiting until the AL language extension is actually needed, which would be the method which would work for both BCCH itself and AL-Go.

MODUSCarstenScholling:
But to implement this properly, I think an additional cache path needs to be implemented for the AL language. To make that work with the symlink and recreate the link after retrieving the data from (compiler) cache. Without the additional cache, I assume there could be inconsistencies or other issues.

I would defer this until we have a better insight into AL-Go development.

PR #3842, opened 2025-02-25
fix: ensure application family is properly used on appsource deployment

warlof:
Currently, the module is not properly forwarding `ApplicationFamily` top/down the stack.
As a result, when we try to deploy an package from the App Source on an environment involving in Embed Application program (where Application Family is needed) - the script crash stating the environment is not available.

This is caused by `$deploymentUrl` which is currently not passing the application family - and as a result - attempt to retrieve an environment using standard Business Central deployment.

Second issue occures while the script attempt to trigger the extension deployment against an environment with `$installUri` which is currently not passing the application family - and as a result - attempt to retrieve an environment using standard Business Central deployment.

Third issue occures while the script attempt to retrieve the extension deployment status against en environment with `$operationUri` which is currently not passing the application family - and as a result - attempt to retrieve an environment using standard Business Central deployment.

This Pull Request is addressing those issues.

### DIFF
--- a/Saas/Install-BcAppFromAppSource.ps1
+++ b/Saas/Install-BcAppFromAppSource.ps1
@@ -1,4 +1,4 @@
-﻿<#
+<#
  .Synopsis
   Function for installing an AppSource App in an online Business Central environment
  .Description
@@ -58,12 +58,16 @@ function Install-BcAppFromAppSource {
         if ($bcEnvironment.Type -eq 'Production' -and !$allowInstallationOnProduction) {
             throw "If you want to install an app in a production environment, you need to specify -allowInstallOnProduction"
         }
-        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
+        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
         if ($appExists) {
             Write-Host -ForegroundColor Green "App $($appExists.Name) from $($appExists.Publisher) version $($appExists.Version) is already installed"
         }
         else {
-            $response = Invoke-RestMethod -Method Get -Uri "$($bcContainerHelperConfig.baseUrl.TrimEnd('/'))/$($bcAuthContext.tenantID)/$environment/deployment/url"
+            $deploymentUrl = "$($bcContainerHelperConfig.baseUrl.TrimEnd('/'))/$($bcAuthContext.tenantID)/$environment/deployment/url"
+            if ($applicationFamily -ne "BusinessCentral") {
+                $deploymentUrl = "https://$($applicationFamily).bc.dynamics.com/$($bcAuthContext.tenantID)/$($environment)/deployment/url"
+            }
+            $response = Invoke-RestMethod -Method Get -Uri $deploymentUrl
             if ($response.status -ne 'Ready') {
                 throw "environment not ready, status is $($response.status)"
             }
@@ -79,7 +83,8 @@ function Install-BcAppFromAppSource {
 
             Write-Host "Installing $appId $appVersion on $($environment)"
             try {
-                $operation = Invoke-RestMethod -Method Post -UseBasicParsing -Uri "$($bcContainerHelperConfig.apiBaseUrl.TrimEnd('/'))/admin/$apiVersion/applications/BusinessCentral/environments/$environment/apps/$appId/install" -Headers $headers -ContentType "application/json" -Body ($body | ConvertTo-Json)
+                $bcAuthContext, $headers, $installUri = Create-SaasUrl -bcAuthContext $bcAuthContext -applicationFamily $applicationFamily -environment $environment -apiVersion $apiVersion -endpoint "apps/$appId/install"
+                $operation = Invoke-RestMethod -Method Post -UseBasicParsing -Uri $installUri -Headers $headers -ContentType "application/json" -Body ($body | ConvertTo-Json)
             }
             catch {
                 throw (GetExtendedErrorMessage $_)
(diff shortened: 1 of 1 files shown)


Saas/Install-BcAppFromAppSource.ps1
```diff
@@ -1,4 +1,4 @@
-﻿<#
+<#
```
freddydk, line 1: there seem to be a change in UTF8 BOM here???

Saas/Install-BcAppFromAppSource.ps1
```diff
@@ -58,12 +58,16 @@ function Install-BcAppFromAppSource {
...
             throw "If you want to install an app in a production environment, you need to specify -allowInstallOnProduction"
         }
-        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
+        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
         if ($appExists) {
             Write-Host -ForegroundColor Green "App $($appExists.Name) from $($appExists.Publisher) version $($appExists.Version) is already installed"
         }
         else {
-            $response = Invoke-RestMethod -Method Get -Uri "$($bcContainerHelperConfig.baseUrl.TrimEnd('/'))/$($bcAuthContext.tenantID)/$environment/deployment/url"
+            $deploymentUrl = "$($bcContainerHelperConfig.baseUrl.TrimEnd('/'))/$($bcAuthContext.tenantID)/$environment/deployment/url"
```
freddydk, line 66: This fix doesn't seem wrong.
In order for containerhelper to work with applicationFamily you need to set baseUrl and ApiBaseUrl in containerhelperconfig - then this fix isn't needed - or???

Saas/Install-BcAppFromAppSource.ps1
```diff
@@ -58,12 +58,16 @@ function Install-BcAppFromAppSource {
         if ($bcEnvironment.Type -eq 'Production' -and !$allowInstallationOnProduction) {
             throw "If you want to install an app in a production environment, you need to specify -allowInstallOnProduction"
         }
-        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
+        $appExists = Get-BcPublishedApps -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily -environment $environment -apiVersion $apiVersion | Where-Object { $_.id -eq $appid -and $_.state -eq "installed" }
```
freddydk, line 61: This fix looks right

freddydk (changes requested):

Saas/Install-BcAppFromAppSource.ps1
warlof: Since we got the `ApplicationFamily` as parameter, it seems more logic to me the method itself handle such changes than altering the configuration wrapper after importing the module.

Otherwise, what's the purpose of having the parameter in the signature?

freddydk: Well - then you need to change that in many places, including inside Get-BcPublishedApps, where apiBaseUrl is used and many more functions. I think we should fix the bug and not try to change more than necessary.

Saas/Install-BcAppFromAppSource.ps1
warlof: Ops, I use GitHub to update the repository. I'll fixed that.

Saas/Install-BcAppFromAppSource.ps1
warlof: so if I get it right, when using the module with Embed Environment, we should do the following - correct ?

```ps
$applicationFamily = "embedPrefix"

Import-Module BcContainerHelper
$bccontainerHelperConfig.baseUrl = "https://$($applicationFamily).bc.dynamics.com"
$bcAuthContext = New-BcAuthContext -ClientId $clientId -ClientSecret $clientSecret -TenantId $tenantId

$environments = Get-BcEnvironments -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily
$installedApplications = Get-BcPublishedApps -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily -environment $environment -apiVersion $apiVersion
$operations = Get-BcEnvironmentOperations -bcAuthContext $bcauthcontext -applicationFamily $applicationFamily -environment $environment
```

Since there is the `Create-SaasUrl` helper method, I assume we shouldn't have to alter the `$bccontainerHelperConfig.ApiBaseUrl` - or am I wrong ?

freddydk: I am also not sure that your suggested fix will work for all - there might be different baseUrls and apiBaseUrl constructions.

freddydk (approved):

Saas/Install-BcAppFromAppSource.ps1
warlof: > To be honest, I cannot remember what you need to set and what not - it has been quite a while since this was added. You might need to set both and use the applicationFamily. The Create-SaasUrl uses the apiBaseUrl - so you probably have to.
> 
> Have you tested this?

It works 👍 
It appears when using admin endpoint, the baseApiUrl can remain the default one.
This is confirmed by the documentation :
- API Admin endpoints pattern : https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/embedapps/embed-app-application-access-management-api
- API endpoints using Embed Application pattern : https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/deployment/embed-app-using-application-family
- API endpoints using standalone pattern : https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/api-reference/v2.0/endpoints-apis-for-dynamics

However, now, the `baseUrl` have to be updated before calling `Install-BcAppFromAppSource` - which also make sense according to our talk from yesterday.

freddydk:
Thanks for the contribution

Issue #3839, opened 2025-02-21
New-BcCompilerFolder populates symbols with different naming pattern, depending on used artifact


MODUSCarstenScholling:
**Describe the issue**
New-BcCompilerFolder populates symbols with different naming pattern, depending on used artifact. This is causing missing symbols when publishing container symbols/apps, because they are not found with pattern *_*.*.*.*.app.

I see three different combinations:

- Sandbox DE artifacts all have versions in their symbol names
- OnPrem DE artifacts don't have any versions at all
- Sandbox W1 has a funny mix, regular symbols with versions, test symbols without.

This is causing errors similar to the following:

```
2025-02-20T14:56:37.6248236Z error AL1024: A package with publisher 'Microsoft', name 'Tests-TestLibraries', and a version compatible with '25.0.0.0' could not be loaded. Symbols for the requested app Tests-TestLibraries by Microsoft 25.0.0.0 could not be found in the database.
2025-02-20T14:56:37.6262209Z error AL1024: A package with publisher 'Microsoft', name 'System Application Test Library', and a version compatible with '25.0.0.0' could not be loaded. Symbols for the requested app System Application Test Library by Microsoft 25.0.0.0 could not be found in the database.
```

**Scripts used to create container and cause the issue**

This should create the mixed version symbols folder:

```PowerShell
$type = 'sandbox';
$version = '26.0.29552.0';
$artifactUrl = "https://bcartifacts-exdbf9fwegejdqak.b02.azurefd.net/$type/$version/w1";

$daFolda = New-BcCompilerFolder -artifactUrl $artifactUrl -containerName "$type-$version"
```

**Additional context**
I have fixed this locally in HelperFunctions.ps1 in function GetTestToolkitApps by collecting the app files with and without version numbers. My first idea to normalize the content of the symbols folder does not work, because that would need repopulation.

```PowerShell
        $appFiles = @()
        $apps | ForEach-Object {
            $tempAppFiles = @(Get-ChildItem -Path $symbolsFolder -Filter "$($_)_*.*.*.*.app");
            
            if ($tempAppFiles.Count -le 0) {
                $tempAppFiles += @(Get-ChildItem -Path $symbolsFolder -Filter "$($_).app");
            }

[...]

PR #3868, opened 2025-03-21
Fixes 3831 and 3839: Local vsix file and Artifacts with symbol files without versions

MODUSCarstenScholling:
This is a reduced version of the former closed PR #3834.

- Allow vsixFile to be a local file for New-BCCompilerFolder (#3831)
- Fixes issues determining test symbols for W1 and OnPrem artifacts not containing versions in their name (#3839)

After that, there is still an issue with RunAlTool always downloading the AL Language extension. There should be a check or an additional parameter as hint for an existing vsix or extracted vsix.

Regarding the other PR: I would recommend modifying the caching paths (in AL-Go) a bit to allow caching of the single vsix in a different base path to not mixing up caching. When specified, this is most likely a different version compared to the rest of the compiler folder contents.


CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -199,13 +199,32 @@ try {
     if ($vsixFile) {
         # If a vsix file was specified unpack directly to compilerfolder
         Write-Host "Using $vsixFile"
-        $tempZip = Join-Path ([System.IO.Path]::GetTempPath()) "alc.$containerName.zip"
-        Download-File -sourceUrl $vsixFile -destinationFile $tempZip
-        Expand-7zipArchive -Path $tempZip -DestinationPath $containerCompilerPath
+
```
freddydk, line 202: Please do not use TAB characters - everything else in BcContainerHelper is spaces.

CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -199,13 +199,32 @@ try {
...
-        Download-File -sourceUrl $vsixFile -destinationFile $tempZip
-        Expand-7zipArchive -Path $tempZip -DestinationPath $containerCompilerPath
+
+		if (Test-Path -Path $vsixFile) {
+			if (Test-Path -Path $vsixFile -PathType Leaf) {
+				Expand-7zipArchive -Path $vsixFile -DestinationPath $containerCompilerPath
+			}
+
+			elseif (Test-Path -Path $vsixFile -PathType Container) {
+				Copy-Item -Path $vsixFile -Destination $containerCompilerPath -Recurse -Force
```
freddydk, line 209: This doesn't copy the content of the vsixFile folder - it includes the folder name in the copy.
I think you need something like:
```
copy-item -Path (Join-Path $vsixFile '*') -Des...
```

HelperFunctions.ps1
```diff
@@ -376,7 +376,7 @@ function GetTestToolkitApps {
         $symbolsFolder = Join-Path $compilerFolder "symbols"
         # Add Test Framework
         $apps = @()
-        $baseAppInfo = Get-AppJsonFromAppFile -appFile (Get-ChildItem -Path $symbolsFolder -Filter 'Microsoft_Base Application_*.*.*.*.app').FullName
+        $baseAppInfo = Get-AppJsonFromAppFile -appFile (Get-ChildItem -Path $symbolsFolder -Filter 'Microsoft_Base Application*.app').FullName
```
freddydk, line 379: Please use the same mechanism as your suggestion below?
i.e. - try to find one with version number and if that doesn't exist grab one without.

With this fix, you might find an app called `Microsoft_Base Application Stress Tests.app` - if something like that would ever exist

freddydk (changes requested):

CompilerFolderHandling/New-BcCompilerFolder.ps1
MODUSCarstenScholling: All tabs converted to 4 spaces

HelperFunctions.ps1
MODUSCarstenScholling: First try find with version pattern *.*.*.*, then without.

CompilerFolderHandling/New-BcCompilerFolder.ps1
MODUSCarstenScholling: Indeed, in my tests the compiler folder was missing.

HelperFunctions.ps1
```diff
@@ -376,7 +376,12 @@ function GetTestToolkitApps {
         $symbolsFolder = Join-Path $compilerFolder "symbols"
         # Add Test Framework
         $apps = @()
-        $baseAppInfo = Get-AppJsonFromAppFile -appFile (Get-ChildItem -Path $symbolsFolder -Filter 'Microsoft_Base Application*.app').FullName
+        $baseAppInfo = Get-AppJsonFromAppFile -appFile (Get-ChildItem -Path $symbolsFolder -Filter 'Microsoft_Base Application_*.*.*.*.app').FullName
```
freddydk, line 379: Isn't this failing with FullName doesn't exist on null pointer?
Don't you need to find the app first and then grab info afterwards?
Like below

MODUSCarstenScholling: You are of course right. I have now added a function and updated Run-TestsInNavContainer to cover it. Tested without issues against OnPrem and Sandbox folder.

freddydk:
FYI - Will revisit this tomorrow (US timezone)

MODUSCarstenScholling:
@freddydk Would you please take another look at this PR?

freddydk:
Will do

CompilerFolderHandling/New-BcCompilerFolder.ps1
```diff
@@ -199,13 +199,36 @@ try {
     if ($vsixFile) {
         # If a vsix file was specified unpack directly to compilerfolder
         Write-Host "Using $vsixFile"
-        $tempZip = Join-Path ([System.IO.Path]::GetTempPath()) "alc.$containerName.zip"
-        Download-File -sourceUrl $vsixFile -destinationFile $tempZip
-        Expand-7zipArchive -Path $tempZip -DestinationPath $containerCompilerPath
+
+        if (Test-Path -Path $vsixFile) {
```
freddydk, line 203:
```suggestion
        if ($vsixFile -notlike 'https://*') {
```

freddydk (changes requested):
Two very small (non functional) things

freddydk:
Could you also resolve conflicts in release notes?

PR #3880, opened 2025-03-31
add excludebuilds

freddydk:
This is NOT the fix for todays problem with the artifacts - this is a fix for future problems of the same kind, allowing people to globally de-select certain builds in ALGoOrgSettings with the stroke of a pen.

Add setting to exclude builds from Get-BcArtifactUrl
The setting ExcludeBuilds is an array of NuGet ranges of builds to exclude.
If you want to exclude a specific build only, use `[27.0.32389.0]`

Ranges are described here: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges

This setting:

```
"ExcludeBuilds": [ "(25.5.30849.32364,25.5.30849.99999)","(26.0.30643.32387,26.0.30643.99999)","(27.0.32389.0,)" ]
```

would remove the affected builds from todays issue

Exclutions from the function when calling get next major:

Excluding 25.5.30849.32390/base
Excluding 26.0.30643.32391/us
Excluding 26.0.30643.32429/us
Excluding 26.0.30643.32442/us
Excluding 27.0.32401.0/us
Excluding 27.0.32402.0/us
Excluding 27.0.32408.0/us
Excluding 27.0.32439.0/us
Excluding 27.0.32446.0/us
https://bcinsider-fvh2ekdjecfjd6gk.b02.azurefd.net/sandbox/26.0.30643.32387/us

The reason for the 25.5 is because it looks for the latest artifacts before determining next major

Another Exclution list (which would work now would be):

```
"ExcludeBuilds": [ "[25.5.30849.32390]","[26.0.30643.32391]","[26.0.30643.32429]","[26.0.30643.32442]","[27.0.32401.0]","[27.0.32402.0]","[27.0.32408.0]","[27.0.32439.0]","[27.0.32446.0]" ]
```

But if another build makes it out before the error is fixed, then that would need to be added too.


freddydk:
The affected builds from today are being deleted so this settings is for the future, if something like this would ever happen again

aholstrup1 (approved):

Issue #3889, opened 2025-04-09
Don't overwrite .nupkg contents if the name is the same


hemisphera:
**Describe the issue**
Using `Download-BcNugetPackageToFolder` overwrites the `.app`-files contained in the `.nupkg` if they have the same name. 

Suggestion: when copying the extracted contents of the main package and all its subpackages, if a file with that name already exists, append the appId to the filename. This avoids overwriting the files, if multiple different packages/dependencies happen to have the same filename within the `.nupkg`.

freddydk:
Fixed in the latest preview

lktraser:
Hi @hemisphera, this is causing some issues for us as we also use NuGet for Runtime packages and want the extra naming schema for those as they are generated in Runtime Conversion example: "TRASER Software GmbH_TRASER Storage_2302.241100.20250425.1_RUNTIME-23.2-w1.app"

It would be better if this renaming of the NuGet package contents were optional. Sadly we still don't have any meta-data in the app package for it's runtime creation version and locale so we only have the name as an option.

PR #3890, opened 2025-04-09
#3889: Use standard app filename for app instead of the app name inside the nuget package

hemisphera:
See #3889 for more info.

For legacy reasons, the name of the .app-Files within our .nupkg packages is the same for all of our apps. We're migrating away from that (for obvious reasons) but in the meantime this would mitigate the problem. Not sure if this causes any trouble downstream.


NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -292,11 +292,18 @@ try {
...
-                Write-Host "Copying $($appFile.Name) to $folder"
-                Copy-Item $appFile.FullName -Destination $folder -Force
+                $TargetFolders = @($folder)
                 if ($copyInstalledAppsToFolder) {
-                    Write-Host "Copying $($appFile.Name) to $copyInstalledAppsToFolder"
-                    Copy-Item $appFile.FullName -Destination $copyInstalledAppsToFolder -Force
+                    $TargetFolders += $copyInstalledAppsToFolder
+                }
+                foreach ($CurrTargetFolder in $TargetFolders) {
+                    $TargetFile = [IO.FileInfo]::new([IO.Path]::Combine($CurrTargetFolder, $appFile.Name))
```
freddydk, line 300: I would prefer using Join-Path, Test-Path and New-Item instead of the .net variants of the same things - this is done all over in BcContainerHelper.
I know it might be a bit more cumbersome, but still.

freddydk: Also, I think there is a problem in just adding the appId - the path can be pretty deep here.
I would prefer something like this:

```
            foreach($appFile in $appFiles) {
                $manifest = Get-AppJsonFromAppFile -appFile $appFile.FullName
                $appFileName = "$($manifest.Publisher)_$($manifest.Name)_$($manifest.Version).app".Split([System.IO.Path]::GetInvalidFileNameChars()) -join ''
                $targetFolders = @($folder)
                if ($copyInstalledAppsToFolder) {
                    $targetFolders += @($copyInstalledAppsToFolder)
                }
                foreach($targetFolder in $targetFolders) {
                    if (-not (Test-Path -Path $targetFolder)) {
                        New-Item -Path $targetFolder -ItemType Directory | Out-Null
                    }
                    Copy-Item -Path $appFile.FullName -Destination (Join-Path $targetFolder $appFileName) -Force
                }
            }
```

freddydk (changes requested):

NuGet/Download-BcNuGetPackageToFolder.ps1
Thomas-Torggler-EOS: Pushed something very similar. Works fine here.

Although I was thinking: we could do without reading the manifest from the app and just simply name the files always as `$(appId).app`. It doesn't require downloading ALC and reading the manifest because we already have the ID at hand. Makes things less readable though. Your call.

hemisphera: Pushed something very similar. Works fine here.

Although I was thinking: we could do without reading the manifest from the app and just simply name the files always as `$(appId).app`. It doesn't require downloading ALC and reading the manifest because we already have the ID at hand. Makes things less readable though. Your call.

freddydk: The ALC will be cached and I like to use the standard app file name pattern whenever I store app files.

NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -291,12 +291,25 @@ try {
...
+                    {
+                        New-Item -Path $CurrTargetFolder -ItemType Directory | Out-Null
+                    }
+                    $appFileName = "$($manifest.Publisher)_$($manifest.Name)_$($manifest.Version)"
+                    $appFileName = $appFileName.Split([System.IO.Path]::GetInvalidFileNameChars()) -join "_"
+                    $TargetFilePath = Join-Path -Path $CurrTargetFolder -ChildPath "$($appFileName).app"
+
+                    Write-Host "Copying $($appFileName) to $CurrTargetFolder"
+                    Copy-Item $appFile.FullName -Destination $TargetFilePath -Force
                 }
```
freddydk, line 313: Some variable casings, formatting and then use the same app file name as everywhere else (avoid replacing invalid characters with \_)

```suggestion
                $targetFolders = @($folder)
                if ($copyInstalledAppsToFolder) {
                    $targetFolders += @($copyInstalledAppsToFolder)
                }
                $manifest = Get-AppJsonFromAppFile -appFile $appFile.FullName
                Write-Host "Found app file for package $($manifest.Name)"
                foreach ($currTargetFolder in $targetFolders) {
                    if (-not (Test-Path -Path $currTargetFolder -PathType Container)) {
                        New-Item -Path $currTargetFolder -ItemType Directory | Out-Null
                    }
                    $appFileName = "$($manifest.Publisher)_$($manifest.Name)_$($manifest.Version)".Split([System.IO.Path]::GetInvalidFileNameChars()) -join ''
                    $targetFilePath = Join-Path -Path $currTargetFolder -ChildPath "$($appFileName).app"

                    Write-Host "Copying $($appFileName) to $currTargetFolder"
                    Copy-Item $appFile.FullName -Destination $targetFilePath -Force
                }
```

freddydk (changes requested):

NuGet/Download-BcNuGetPackageToFolder.ps1
lukas-pfeifhofer: For me this solution works.

freddydk (approved):

PR #3892, opened 2025-04-11
Use strong file name pattern when downloading NuGet packages

lukas-pfeifhofer:
For some legacy scenarios our NuGet packages always use the same file name for the `*.app` files. This will be a problem by using the `Download-BcNuGetPackageToFolder` cmdlt, as the latest downloaded `*.app` file overwrites the previous one, especially when using AL-Go with 3rd-party feeds (see also #3889).

Before copying the `*.app` file to the destination folder, the file name is checked for a simple name pattern, if this is the case then the destination file name is converted with a stronger name pattern to avoid overwritting.


freddydk:
I have added a suggestion to https://github.com/microsoft/navcontainerhelper/pull/3890 with the way I think this should be solved. You can try this to see whether it solves the issue for you?

freddydk:
We will go with the fix suggested in https://github.com/microsoft/navcontainerhelper/pull/3890 as soon as that is approved.

PR #3901, opened 2025-04-15
Improve performance of NuGet Download

OleWunschmann:
The following example results in a long runtime caused by unnecessary app downloads and high number of web requests:
```powershell
Import-Module BcContainerHelper -Force -DisableNameChecking
$bcContainerHelperConfig.TrustedNuGetFeeds = @()
$bcContainerHelperConfig.TrustedNuGetFeeds += @([PSCustomObject]@{ 'Url' = 'https://dynamicssmb2.pkgs.visualstudio.com/DynamicsBCPublicFeeds/_packaging/MSApps/nuget/v3/index.json'; 'Token' = ''; 'Patterns' = @('*'); 'Fingerprints' = @() })
$bcContainerHelperConfig.TrustedNuGetFeeds += @([PSCustomObject]@{ 'Url' = 'https://dynamicssmb2.pkgs.visualstudio.com/DynamicsBCPublicFeeds/_packaging/MSSymbols/nuget/v3/index.json'; 'Token' = ''; 'Patterns' = @('*'); 'Fingerprints' = @() })
$bcContainerHelperConfig.TrustedNuGetFeeds += @([PSCustomObject]@{ 'Url' = 'https://dynamicssmb2.pkgs.visualstudio.com/DynamicsBCPublicFeeds/_packaging/AppSourceSymbols/nuget/v3/index.json'; 'Token' = ''; 'Patterns' = @('*'); 'Fingerprints' = @() })
Measure-Command {
        Download-BcNuGetPackageToFolder -packageName Microsoft.Application -folder C:\Temp\Test -downloadDependencies all -select 'LatestMatching' -installedPlatform '25.0'
} | Format-Table
```

This PR introduces the following improvements:
- Reduce the number of downloads by:
  - Only retrieving the nuspec to check the package and its dependencies.
  - Checking all direct dependencies before downloading (both the package and its dependencies).
- Reduce the number of web requests by caching the search results for a package name per feed (for 10 minutes).


OleWunschmann:
@microsoft-github-policy-service agree company="Cosmo Consult"

NuGet/NuGetFeedClass.ps1
```diff
@@ -83,7 +85,20 @@ class NuGetFeed {
     }
 
     [hashtable[]] Search([string] $packageName) {
-        if ($this.searchQueryServiceUrl -match '^https://nuget.pkg.github.com/(.*)/query$') {
+        $useCache = $this.searchResultsCache.ContainsKey($packageName)
+        if ($useCache) {
+            # Clear cache older than 10 minutes
+            if ($this.searchResultsCache[$packageName].timestamp -lt (Get-Date).AddMinutes(-10)) {
```
freddydk, line 91: Please add the cache retention period to the bcContainerHelper settings in bc.helperfunctions.ps1 - this allows people to control this (or remove it) if needed - and make sure that 0 means do not use cache.

freddydk (changes requested):
This looks really good - a few minor comments.
Also, please add a few lines in the release notes specifying what is done and the new setting (when that is done) and the default value of that.

NuGet/NuGetFeedClass.ps1
```diff
@@ -83,7 +85,20 @@ class NuGetFeed {
     }
 
     [hashtable[]] Search([string] $packageName) {
-        if ($this.searchQueryServiceUrl -match '^https://nuget.pkg.github.com/(.*)/query$') {
+        $useCache = $this.searchResultsCache.ContainsKey($packageName)
```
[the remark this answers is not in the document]
OleWunschmann: Good point. 
My suggestion would be to remove the whole cache to keep it simple.
Another solution would be to only remove items with keys that match the pushed package name (`$packageName -like "*$Key*"`).
Which option would you prefer?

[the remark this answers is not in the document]
freddydk: The latter - running with incremental builds, we will publish packages if they don't already exist - it would be good that the cache isn't cleared all the time (even though it might not matter at all)

NuGet/NuGetFeedClass.ps1
```diff
@@ -85,16 +85,31 @@ class NuGetFeed {
     }
 
     [hashtable[]] Search([string] $packageName) {
-        $useCache = $this.searchResultsCache.ContainsKey($packageName)
-        if ($useCache) {
-            # Clear cache older than 10 minutes
-            if ($this.searchResultsCache[$packageName].timestamp -lt (Get-Date).AddMinutes(-10)) {
+        $cacheRetentionPeriod = 600 # 10 minutes
```
freddydk, line 88: I would prefer to have a class variable like:

[int] $cacheRetentionPeriod = $bcContainerHelperConfig.NuGetSearchResultsCacheRetentionPeriod

and then use $this.cacheRetentionPeriod in the search function (avoiding the two signatures of search)

freddydk (changes requested):

NuGet/NuGetFeedClass.ps1
OleWunschmann: I will remove the second signiture.

I also thought about a class variable but the current solution allows the change of the retention period between searches.
But I can change it if you still prefer the class variable.

freddydk: My problem was really that I had to test what $script: would do in this case - I find it harder to read.
Maybe that's just me:-)
In the end - the setting is only for allowing people to add a setting in AL-Go to change this / disable this if things stops working for some reason.

OleWunschmann: I thought about it and made your suggested change.

My original implementation had no config and always cached for 10 minutes.
So nothing is lost with your suggestion and it keeps it simpler and - as you mentioned - easier to read.

freddydk (approved):

PR #3910, opened 2025-04-24
Don't remove packagesFolder if it already exists

PatrickSchiefer:
If compilefolders are used the packages folder will be deleted and recreated during the pipeline run.

In certain cases especially when you use customized AL:Go with foreign container systems, this folder exists on purpose and already includes files. This files should not be deleted because it will lead the pipeline to fail

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -875,10 +875,9 @@ if ($updateLaunchJson) {
 
 if ($useCompilerFolder -or $filesOnly -or !$useDevEndpoint) {
     $packagesFolder = CheckRelativePath -baseFolder $baseFolder -sharedFolder $sharedFolder -path $packagesFolder -name "packagesFolder"
-    if (Test-Path $packagesFolder) {
-        Remove-Item $packagesFolder -Recurse -Force
+    if (!(Test-Path $packagesFolder)) {
+        New-Item $packagesFolder -ItemType Directory | Out-Null
     }
-    New-Item $packagesFolder -ItemType Directory | Out-Null
 }
 
 if ($useDevEndpoint) {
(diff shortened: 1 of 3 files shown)


AppHandling/Run-AlPipeline.ps1
```diff
@@ -875,9 +875,12 @@ if ($updateLaunchJson) {
...
 if ($useCompilerFolder -or $filesOnly -or !$useDevEndpoint) {
     $packagesFolder = CheckRelativePath -baseFolder $baseFolder -sharedFolder $sharedFolder -path $packagesFolder -name "packagesFolder"
-    if (!(Test-Path $packagesFolder)) {
-        New-Item $packagesFolder -ItemType Directory | Out-Null
+    if (!($bcContainerHelperConfig.doNotRemovePackagesFolderIfExists)) {
+        if (Test-Path $packagesFolder) {
+            Remove-Item $packagesFolder -Recurse -Force
+        }
     }
+    New-Item $packagesFolder -ItemType Directory -Force | Out-Null
```
freddydk, line 883: If the folder already exists - won't this fail?

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
freddydk: I think this needs to be something like::
```
    if (Test-Path $packagesFolder) {
        if (!($bcContainerHelperConfig.doNotRemovePackagesFolderIfExists)) {
            Remove-Item $packagesFolder -Recurse -Force
            New-Item $packagesFolder -ItemType Directory -Force | Out-Null 
        }
    }
    else {
        New-Item $packagesFolder -ItemType Directory -Force | Out-Null 
    }
```

freddydk (changes requested):

AppHandling/Run-AlPipeline.ps1
PatrickSchiefer: No according to the documentation the force will not fail for directories. The behavavior is that it will create if the folder exists and skip if it not exists

See documentation of New-Item here
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-7.5#example-8-use-the-force-parameter-to-attempt-to-recreate-folders

freddydk: Got it - and that also works on 5.1 - thanks

freddydk (approved):

PR #3925, opened 2025-05-06
Limit assembly probing paths to a specific dotnet version when compiling in a container 

mazhelez:
Limit assembly probing paths to a specific dotnet version when compiling in a container


freddydk:
I think we should do @aholstrup1 's suggestion

PR #3937, opened 2025-05-13
Add cache control during releases

aholstrup1:
When releasing bccontainerhelper we upload the new version to a storage account. When these blobs are accessed through AFD we should specify what the cache duration should be so AFD takes that into consideration.

### DIFF
--- a/.github/workflows/Release.yaml
+++ b/.github/workflows/Release.yaml
@@ -108,9 +108,8 @@ jobs:
             New-AzStorageContainer -Name 'public' -Context $storageContext -Permission 'Container' -ErrorAction Ignore | Out-Null
           
             Compress-Archive -path $path -DestinationPath "$($path).zip"
-            $cacheDurationSeconds = 3600 # 1 hour
             Set-AzStorageBlobContent -File "$($path).zip" -Context $storageContext -Container 'public' -Blob "$version.zip" -Force | Out-Null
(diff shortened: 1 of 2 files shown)


.github/workflows/Release.yaml
```diff
@@ -108,8 +108,9 @@ jobs:
             New-AzStorageContainer -Name 'public' -Context $storageContext -Permission 'Container' -ErrorAction Ignore | Out-Null
           
             Compress-Archive -path $path -DestinationPath "$($path).zip"
+            $cacheDurationSeconds = 3600 # 1 hour
             Set-AzStorageBlobContent -File "$($path).zip" -Context $storageContext -Container 'public' -Blob "$version.zip" -Force | Out-Null
-            Set-AzStorageBlobContent -File "$($path).zip" -Context $storageContext -Container 'public' -Blob "latest.zip" -Force | Out-Null
+            Set-AzStorageBlobContent -File "$($path).zip" -Context $storageContext -Container 'public' -Blob "latest.zip" -Properties @{"CacheControl" = "max-age=$cacheDurationSeconds"} -Force | Out-Null
```
freddydk, line 113: I think we should do no-cache on both.
Reason being that we would need to always remember to release a new BcContainerHelper 1 hour before we release AL-Go, because you could have the old version cached for an hour.

freddydk (changes requested):

freddydk (approved):

PR #4041, opened 2025-11-10
NuGet performance

freddydk:
Better performance when downloading NuGet packages by adding another cache folder, which defaults to c:\bcnuget.cache
Add bcnuget to cache parameter in Flush-ContainerHelperCache in order to cleanup NuGet cache folders as well
Better performance when searching for matching packages by not re-searching multiple times, but instead fetching all versions just once
Better performance when reading nuspec metadata - being cached as well.

TODO
- [x] Cache package metadata in the class
- [x] Docs

### DIFF
--- a/NuGet/Push-BcNuGetPackage.ps1
+++ b/NuGet/Push-BcNuGetPackage.ps1
@@ -23,7 +23,8 @@ Function Push-BcNuGetPackage {
         [string] $bcNuGetPackage
     )
 
-    $nuGetFeed = [NuGetFeed]::Create($nuGetServerUrl, $nuGetToken, @(), @(),$bcContainerHelperConfig.NuGetSearchResultsCacheRetentionPeriod,$bcContainerHelperConfig.BcNuGetCacheFolder)
+    $nuGetFeed = [NuGetFeed]::Create($nuGetServerUrl, $nuGetToken, @(), @(), $bcContainerHelperConfig.NuGetSearchResultsCacheRetentionPeriod, $bcContainerHelperConfig.BcNuGetCacheFolder)
+
     $nuGetFeed.PushPackage($bcNuGetPackage)
 }
 Export-ModuleMember -Function Push-BcNuGetPackage


NuGet/Find-BcNuGetPackage.ps1
```diff
@@ -46,11 +46,12 @@ Function Find-BcNuGetPackage {
         [Parameter(Mandatory=$false)]
         [string[]] $excludeVersions = @(),
         [Parameter(Mandatory=$false)]
-        [ValidateSet('Earliest','Latest','Exact','Any')]
+        [ValidateSet('Earliest','Latest','Exact','Any','AllAscending','AllDescending')]
```
mazhelez: Doc is missing.

NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -135,9 +135,18 @@ try {
...
+    }
+    foreach($nuGetVersion in $nuGetVersions) {
+        $feed = $nuGetVersion.feed
+        $packageId = $nuGetVersion.packageId
+        $packageVersion = $nuGetVersion.packageVersion
         $returnValue = @()
-        $feed, $packageId, $packageVersion = Find-BcNugetPackage -nuGetServerUrl $nuGetServerUrl -nuGetToken $nuGetToken -packageName $packageName -version $version -excludeVersions $excludeVersions -verbose:($VerbosePreference -eq 'Continue') -select $findSelect -allowPrerelease:($allowPrerelease.IsPresent)
         if (-not $feed) {
             Write-Host "No package found matching package name $($packageName) Version $($version)"
             break
```
mazhelez: With the new design, does it make sense to break here?

NuGet/Download-BcNuGetPackageToFolder.ps1
mazhelez: I believe `$excludeVersions` becomes redundant, as there won't another call to `Find-BcNugetPackage`

mazhelez (approved):

NuGet/Download-BcNuGetPackageToFolder.ps1
```diff
@@ -80,10 +80,10 @@ Function Download-BcNuGetPackageToFolder {
 try {
     $findSelect = $select
     if ($select -eq 'LatestMatching') {
-        $findSelect = 'Latest'
+        $findSelect = 'AllDescending'
```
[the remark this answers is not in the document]
freddydk: Correct - and the old mechanism also only got one version and then downloaded the metadata afterwards - in order to continue after that...

NuGet/Download-BcNuGetPackageToFolder.ps1
mazhelez: I wouldn't have left the comment otherwise 😉 
But it's okay if it stays.

PR #4053, opened 2025-11-19
Fix: Endless loop when errors occur during the execution of AL tests with external environments

### LINKED ISSUE #3986: [Bug]: Endless loop in PR build in not tests in a Test Codeunit.
### AL-Go version

How do I find it?

### Describe the issue

If you create a Codeunit of `Subtype = Test;` and have no `[Test]` decorated procedures in the codeunit the test process just seems to loop repeatedly.

See attached log file of PR build that failed with this condition.

Admittedly this is an edge case but no condition should cause the process to loop endlessly.  It burned 6 hours of compute time under this condition.

### Expected behavior

Admittedly this is an edge case but no condition should cause the process to loop endlessly.  It burned 6 hours of compute time under this condition.

### Steps to reproduce

Create an AL-GO project
Create a branch
Create a Test Codeunit, 
do NOT add any Test procedures
submit a Pull Request  for the branch

### Additional context (logs, screenshots, etc.)

[logs_41813517224.zip](https://github.com/user-attachments/files/21297122/logs_41813517224.zip)

OleWunschmann:
- Error when running AL tests after the first failed attempt with an external environment or after retry with container
- Handling of test results without the “testResults” property (e.g., if a test codeunit does not contain a test function)
  - AL-Go ran into an error due to the strict mode

Fixes #3986

### DIFF
--- a/.azuredevops/pipelines/Release.yml
+++ b/.azuredevops/pipelines/Release.yml
@@ -1,0 +1,127 @@
+trigger: none
+
+pr: none
+
+pool:
+  vmImage: 'windows-latest'
+
+variables:
+  - group: Release-Secrets # Create a variable group with: AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_SUBSCRIPTION_ID, CodeSignCertificatePfx, CodeSignCertificatePassword, NugetKey
+
+jobs:
+- job: Deploy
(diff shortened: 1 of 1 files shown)


freddykristiansen (approved):
Looks good IMO

PR #4075, opened 2025-12-18
feat: Get-BCArtifactUrl - Add Support for -before and -after Parameters in NextMinor and NextMajor

DanielGoehler:
## Description

This pull request introduces enhanced handling of the `-before` and `-after` parameters when using the `-select NextMinor` or `-select NextMajor` options in the `Get-BCArtifactUrl` cmdlet. Previously, these parameters were not considered when filtering artifact URLs for NextMinor or NextMajor versions.

## Motivation

In our Azure DevOps pipelines—using ALOps—we often need to create containers based on `NextMinor` or `NextMajor` builds. Sometimes, multiple new builds are released on the same day, which can result in pipelines retrieving the latest build of today rather than the last stable build from the day before. This is not inherently problematic, but it can increase pipeline runtimes and introduce instability if today's docker container builds are frequently refreshed.

By supporting the `-before` and `-after` parameters in conjunction with `-select NextMinor` and `-select NextMajor`, we can exclude today's latest build if necessary, and reliably fetch the last build from yesterday or a specified date range. This change should significantly improve pipeline performance and reliability.

## Example Usage

```powershell
Get-BCArtifactUrl -accept_insiderEula -before 2025-12-18 -country de -select NextMinor -type Sandbox
```

This command will fetch the last `NextMinor` build before 2025-12-18 for Germany (`de`), allowing pipelines to intentionally skip today’s builds and use the previous stable release.

## Summary of Changes

- Updated logic to consider `-before` and `-after` with `-select NextMinor`/`NextMajor`
- Improved filtering of builds to support date-based exclusions/inclusions

## Benefits

- More control over artifact selection in automated pipelines
- Reduces unnecessary container creation and pipeline runtime
- Increases reliability by allowing exclusion of freshly published builds when needed
- Over the years, this reduces energy consumption (less build agents needed) and helps reduce the environment impact.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -2,7 +2,7 @@
 Temporarily avoid using dotnet 10 for assemblyProbingPaths
 Issue 3986 Endless loop when errors occur during the execution of AL tests with external environments
 Issue 3986 Execution of AL tests fails for test codeunits without test functions
-
+Get-BCArtifactUrl: Add Support for -before and -after Parameters in NextMinor and NextMajor
 
 6.1.10
 Better performance when downloading NuGet packages by adding another cache folder, which defaults to c:\bcnuget.cache


DanielGoehler:
@mazhelez Thanks. Done.

Issue #4046, opened 2025-11-14
How to unpublish apps after publishing with Publish-PerTenantExtensionApps.ps1


Mkublbock:
We are using Publish-PerTenantExtensionApps.ps1 for our CD pipeline to automatically publish in customer environments. However, after running this script the old app still stays published but not installed. After some time this leads to alot of published apps that are not installed.

How can we add a step to unpublish the upgraded app after running Publish-PerTenantExtensionApps.ps1?

Issue #2808, opened 2022-11-30
Ability to unpublish specific version number of an extension in a SaaS environment


scm6iwa:
This isn't an issue but more of a request.

Could you create a Unpublish-PerTenantExtensionApps script where you can unpublish specific versions of an extension in a BCSaaS environment.

This would be useful so when I publish an extension using Publish-PerTenantExtensionApps to my SaaS sandboxes via my pipeline's I can then also unpublish the uninstalled extensions to keep the sandbox environments clean of none installed extensions.

DavidFeldhoff:
Hi Freddy,
I think my requests relates to this issue: I'd like to uninstall some dev-apps from a SaaS environment.
`Publish-BCContainerApp` e.g. works fine with with the parameters `bcAuthContext` and `environment`, but it seems that `Unpbulish-BcContainerApp` is missing these parameters. Is that doable?

DavidFeldhoff:
The alternative would be to uninstall apps via the BC Admin API. But unfortunately that isn't working for dev-extensions.
@freddydk can you maybe give a short statement if you would pick this up? 
Otherwise I'd create a support request for the BC Admin API that the uninstall endpoint
`https://api.businesscentral.dynamics.com/admin/v2.11/applications/BusinessCentral/environments/$environmentName/apps/$appIdToUninstall/uninstall`
should work for dev-extensions as well, but I could assume that they'd refuse it.

freddydk:
I will investigate what can be done.

DavidFeldhoff:
Awesome, thanks!

MikelSeron:
Hello Freddy!
Do you have any news regarding this topic?
Thanks!

Spycoclown:
Hello @freddydk, are there already some news?

peterlinnet:
At a first glance:
The automation API does not offer a bound action to unpublish extensions.
The admin center API only lists AppSource or firstparty apps, so Dev scope or PTEs are not available here.

So if you use the BC-container-helper to automate updating PTEs through a deployment pipeline, you'll soon have a history of old versions of the PTE cluttering the customer's list of installed (published) extensions:
![image](https://github.com/microsoft/navcontainerhelper/assets/25705087/44d68239-9bc3-426b-8aaf-acd8cee4567d)

I guess this was the intention of the OP and it's the reason why I'm here looking :)

Tedas1:
Hello, did someone solve or found a way to unpublish those cluttered PTE apps?

[...]

PR #4080, opened 2026-01-07
feat: Add support to unpublish old and upgraded apps published with Publish-PerTenantExtensionApps

jniebrueggeSG:
### Summary
This PR adds a new optional switch parameter unpublishOldVersions to the Publish-PerTenantExtensionApps function that automatically unpublishes old versions of apps after successfully upgrading to a new version. This is mentioned in #2808 and #4046.

### Changes
Added unpublishOldVersions switch parameter to function signature
Implemented logic to unpublish old app versions using the Microsoft.NAV.unpublish API endpoint
Added version check to ensure the unpublish API is only called on Business Central versions 25.4.0.0 and later (when the new unpublish API became available)
Updated function documentation to include the new parameter

### Technical Details
The unpublish operation only occurs when:
The unpublishOldVersions switch is explicitly provided
An app is being upgraded (not newly installed)
The Business Central version is 25.4.0.0 or later
The upgrade operation completes successfully
Uses the Microsoft.NAV.unpublish API endpoint available in the automation API

### Breaking Changes
None - this is an additive change that maintains full backward compatibility.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -3,6 +3,7 @@ Temporarily avoid using dotnet 10 for assemblyProbingPaths
 Issue 3986 Endless loop when errors occur during the execution of AL tests with external environments
 Issue 3986 Execution of AL tests fails for test codeunits without test functions
 Get-BCArtifactUrl: Add Support for -before and -after Parameters in NextMinor and NextMajor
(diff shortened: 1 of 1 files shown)


aholstrup1 (approved):

mazhelez:
Can you also please add release notes?

PR #4098, opened 2026-02-11
Improved error message for system symbols

tsr-kmanson:
**Improve System Symbols Error Messages in Compile-AppWithBcCompilerFolder**

**Summary**

Enhanced error handling and messaging when validating system symbols during AL app compilation to provide clearer diagnostics and actionable guidance.

The previous implementation used a generic error message "Unable to locate system symbols" for all failure cases, making it difficult to diagnose whether:

- No system symbols were found at all
- Multiple conflicting system symbol files existed in the symbols folder

**Solution**

Replaced the generic error with conditional error messages that:

- When no system symbols found (Count = 0):
-- Clearly states that no System.app file was found
-- Provides the symbols folder path for verification
- When multiple system symbols found (Count > 1):
-- Explicitly states that only one System.app is expected
-- Shows the actual count of instances found
-- Instructs the user to cleanup the symbols folder
-- Provides the symbols folder path for cleanup

### DIFF
--- a/CompilerFolderHandling/Compile-AppWithBcCompilerFolder.ps1
+++ b/CompilerFolderHandling/Compile-AppWithBcCompilerFolder.ps1
@@ -69,20 +69,20 @@
 #>
 function Compile-AppWithBcCompilerFolder {
     Param (
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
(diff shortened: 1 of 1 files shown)


aholstrup1 (approved):

tsr-kmanson:

Issue #2158, opened 2021-10-29
BC 19.0 Artifacts => WebClient returns "500 - Internal server error"


gregoral:
A newly created  W1 BC container does not work.
WebClient returns: "500 - Internal server error"
![image](https://user-images.githubusercontent.com/42441688/139395146-02c6172f-1124-4589-bb3a-7a6a2554de6a.png)

**Steps taken:**
- create a new image from artifacts
- create a new container from image
- open web client

**Observations:**
- certificate download works fine
- bc server appears to be working ( Get-NAVAppInfo, Publish-NAVApp, Install-NAVApp )
- webclient simply returns error 500
- version 18.5 works fine
- version 19.0 does not
- windows error log does not appear to have anything unusual ( I can provide an evtx export if needed )
- using -isolation hyperv doesn't help

```
# 19.0 returns error 500 when opening webclient
New-BcImage -artifactUrl 'https://bcartifacts.azureedge.net/sandbox/19.0.29894.31767/w1'
docker run -d ... local-bc-image:sandbox-19.0.29894.31767-w1

# 18.5 works fine
New-BcImage -artifactUrl 'https://bcartifacts.azureedge.net/sandbox/18.5.29545.31744/w1'
docker run -d ... local-bc-image:sandbox-18.5.29545.31744-w1 
```

**BcContainerHelper command used to create the image**
```
New-BcImage `
   -artifactUrl 'https://bcartifacts.azureedge.net/sandbox/19.0.29894.31767/w1' `
   -imageName 'local-bc-image' `
   -isolation 'process' `
   -baseImage 'mcr.microsoft.com/dynamicsnav:10.0.17763.1577-generic' `
   -memory '8gb' `
   -skipDatabase:$false `
   -multitenant:$false `
   -addFontsFromPath 'C:\BcFonts' `
   -licenseFile 'C:\bc19.flf' `
   -includeTestToolkit:$true `
   -includeTestFrameworkOnly:$true `
   -includeTestLibrariesOnly:$true `
   -includePerformanceToolkit:$true
```

**Additional context**
```
BcContainerHelper is version 2.0.16
Using -isolation hyperv doesn't help.

local-bc-image:sandbox-19.0.29894.31767-w1
Building image local-bc-image:sandbox-19.0.29894.31767-w1 based on mcr.microsoft.com/dynamicsnav:10.0.17763.1577-generic with https://bcartifacts.azureedge.net/sandbox/19.0.29894.31767/w1
[...]

PR #4143, opened 2026-04-30
Fix BCPT test job failures not surfaced and empty results not reported

jeffreybulanadi:
## Problem

When running BCPT performance tests via the \connectFromHost = \True\ code path, \Run-BCPTTestsInBcContainer.ps1\ launches a background job to run \RunBCPTTests.ps1\ from the host. If that job fails (state is not \Completed\), the current code only prints a message and continues, fetching what is effectively an empty \cptLogEntries\ response from the Business Central API.

The empty result is then saved to the results file. When \AnalyzeTests\ in AL-Go reads the file, it finds no entries, writes nothing to the GitHub step summary, and the build succeeds without any indication that the performance tests did not produce output.

See: microsoft/AL-Go#2158

### Run-BCPTTestsInBcContainer.ps1

- When the background test job does not complete successfully, collect the job output and throw an error. This surfaces the failure immediately rather than silently continuing with an empty result set.
- After fetching \cptLogEntries\, check whether the response contains any entries. If it does not, emit a GitHub Actions warning annotation so the empty result is visible in the build log even when the job exits without error.

## Testing

1. Set up a Business Central container with a BCPT test app.
2. Configure a \cptSuiteCode\ that references a codeunit that does not exist in the container.
3. Run \Run-BCPTTestsInBcContainer\ with the broken suite.
4. Before this fix: the build succeeds with no output.
5. After this fix: a warning annotation appears in the build log stating that no BCPT log entries were found.

For the job failure case:
1. Simulate a job failure by passing invalid parameters that cause the job to fault.
2. Before this fix: a message is printed and execution continues.
3. After this fix: an error is thrown and the build fails with the job output included.

### DIFF
--- a/AppHandling/Run-BCPTTestsInBcContainer.ps1
+++ b/AppHandling/Run-BCPTTestsInBcContainer.ps1
@@ -188,8 +188,9 @@ try {
         } -ArgumentList (Join-Path $bcContainerHelperConfig.hostHelperFolder "extensions\$containerName\my\TestRunner"), $params, $suiteCode, $serviceUrl, $testPage | Wait-Job
 
         if ($job.State -ne "Completed") {
-            Write-Host "Running performance test failed"
-            $job | Receive-Job
+            $jobOutput = $job | Receive-Job | Out-String
+            $job | Remove-Job | Out-Null
+            throw "Running BCPT performance test failed. Job state: $($job.State). Output: $jobOutput"
         }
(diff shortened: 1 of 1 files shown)


jeffreybulanadi:
Closing this PR as the fix for AL-Go issue #2158 is handled in the AL-Go repository directly. The companion navcontainerhelper change is out of scope for this issue.

PR #4145, opened 2026-04-30
Add retry with exponential backoff for transient docker pull failures

mazhelez:
## Problem

Docker pull operations intermittently fail with errors like:

```
Error response from daemon: pull access denied for mcr.microsoft.com/businesscentral, 
repository does not exist or may require 'docker login': denied: <!DOCTYPE html>...
<h2>The request is blocked.</h2>...
```

This is caused by Azure Front Door (the CDN in front of MCR) occasionally blocking requests, returning an HTML error page instead of a proper registry response. Docker misinterprets this as a "pull access denied" error.

## Fix

Added retry logic in `DockerDo` specifically for `pull` commands when a transient CDN/WAF block is detected:

- **Up to 5 retries** (6 total attempts)
- **Exponential backoff**: 2, 4, 8, 16, 32 seconds between attempts
- **Detection heuristic**: error output contains `<html` or `The request is blocked`
- **Non-transient errors** (authentication failures, genuinely missing images) are **not** retried and fail immediately as before
- **Non-pull commands** (`run`, `start`, `build`, etc.) are completely unaffected

## Changes

- `HelperFunctions.ps1`: Wrapped docker process execution in a retry loop for pull commands
- `ReleaseNotes.txt`: Added entry under 6.1.14


jonasblunck:
## Review Feedback

**[Important | Correctness] Retry heuristic is too broad — will retry non-transient failures**

**File**: `HelperFunctions.ps1:249`

The current retry condition:
`powershell
if ($attempt -le $maxRetries -and $err -match 'pull access denied') {
`

This matches **all** `pull access denied` errors — including legitimate ones where the image genuinely doesn't exist or the user actually needs to authenticate. In those cases, users will wait through 5 retries (62 seconds of delay) before seeing the real error.

The PR description states the detection heuristic should check for `<html` or `The request is blocked` (the Azure Front Door HTML response), but the implementation doesn't include that check.

**Impact**: This also means the helpful `You need to login to...` message (line 266) is suppressed during retries, since the retry check at line 249 fires before the `authentication required` check at line 266.

**Suggested fix** — tighten the condition to only match the CDN/WAF block pattern:
`powershell
if ($attempt -le $maxRetries -and $err -match 'pull access denied' -and ($err -match '<html' -or $err -match 'The request is blocked')) {
`

This ensures:
- ✅ Transient CDN/WAF blocks → retried (they contain HTML in the error)
- ✅ Genuinely missing images → fail immediately (no HTML in error)
- ✅ Authentication failures → show login message immediately (no HTML in error)

---

*Minor suggestions (non-blocking):*
- Consider adding `\$p.Dispose()` after `WaitForExit()` to clean up process handles on retry iterations
- The retry message could say "CDN/WAF block" instead of "transient error" to be more informative for users

mazhelez:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/navcontainerhelper/pull/4145#pullrequestreview-4226225121)

ReleaseNotes.txt
```diff
@@ -1,10 +1,11 @@
 6.1.15
 
 6.1.14
+Add retry with exponential backoff in DockerDo for transient MCR pull failures caused by Azure Front Door CDN/WAF blocks
```
mazhelez: @copilot can you please move the release notes under "6.1.15"?

spetersenms (dismissed):

aholstrup1 (dismissed):

mazhelez:
@copilot When testing the fix, the build fails with an error: https://github.com/microsoft/BCApps/actions/runs/25427539970/job/74640939706

Please investigate and create a fix. Do not commit, as I'd like to review.

spetersenms (approved):

Issue #4126, opened 2026-04-22
[Bug]: Get-AppSourceProduct Fails with Bad Request (ICH50400)


PeterDurrer:
### BCContainerhelper version

6.1.12

### Describe the issue
Get-AppSourceProduct is broken and return this Error:
```
Der Remoteserver hat einen Fehler zurückgegeben: (400) Ungültige Anforderung.
Bad Request
"ICH50400 - Unable to complete request. {\"faultCode\":50400,\"reasonPhrase\":\"Request validation failed.\",\"correlat
ionId\":\"70d9822c-6a89-4995-a802-472b7376e74b\",\"operationId\":\"87f621fe67996ec44b04b9d92bc20543\",\"data\":{\"Produ
ctTypeFilter[0]\":[\"The field is invalid (value 'Application').\"],\"ProductTypeFilter[1]\":[\"The field is invalid
(value 'Bundle').\"],\"ProductTypeFilter[2]\":[\"The field is invalid (value
'AvatarItem').\"],\"ProductTypeFilter[3]\":[\"The field is invalid (value 'InternetOfThings').\"]}}"
In C:\Program Files\WindowsPowerShell\Modules\BcContainerHelper\6.1.12\AppSource\Invoke-IngestionAPI.ps1:87 Zeichen:21
+                     throw (GetExtendedErrorMessage $_)
+                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Der Remoteserve...Things').\"]}}":String) [], RuntimeException
    + FullyQualifiedErrorId : Der Remoteserver hat einen Fehler zurückgegeben: (400) Ungültige Anforderung.
Bad Request
    "ICH50400 - Unable to complete request. {\"faultCode\":50400,\"reasonPhrase\":\"Request validation failed.\",\"cor
   relationId\":\"70d9822c-6a89-4995-a802-472b7376e74b\",\"operationId\":\"87f621fe67996ec44b04b9d92bc20543\",\"data\
  ":{\"ProductTypeFilter[0]\":[\"The field is invalid (value 'Application').\"],\"ProductTypeFilter[1]\":[\"The fiel
 d is invalid (value 'Bundle').\"],\"ProductTypeFilter[2]\":[\"The field is invalid (value 'AvatarItem').\"],\"Prod
uctTypeFilter[3]\":[\"The field is invalid (value 'InternetOfThings').\"]}}"

```
[...]

PR #4146, opened 2026-04-30
Fix Get-AppSourceProduct: Add explicit resourceType filter to avoid ICH50400 Bad Request

mmenninghaus:
#### Problem

`Get-AppSourceProduct` fails with a `400 Bad Request` error when calling the `/products` collection endpoint:

```
ICH50400 - Unable to complete request. Request validation failed.
ProductTypeFilter[0]: The field is invalid (value 'Application').
ProductTypeFilter[1]: The field is invalid (value 'Bundle').
ProductTypeFilter[2]: The field is invalid (value 'AvatarItem').
ProductTypeFilter[3]: The field is invalid (value 'InternetOfThings').
```

This is caused by a server-side change to the Partner Center Ingestion API that now rejects the default `ProductTypeFilter` values it previously applied internally.

Related: #4126

#### Fix

Added an explicit `$filter=resourceType eq 'AzureDynamics365BusinessCentral'` query parameter to the `/products` collection calls. This overrides the broken server-side default filter and ensures only Business Central products are returned.

The single-product endpoint (`/products/{id}`) is unaffected and remains unchanged.

#### Changes

- Get-AppSourceProduct.ps1: Pass `$productsQuery` with OData `$filter` to `Invoke-IngestionApiGetCollection` when listing products (both for `-productName` and the no-parameter case).

### ✅ Checklist

- [x] Add tests (`Tests` / `LinuxTests`)
 - Verified that `Get-AppSourceProduct -authContext $ctx -silent` returns products successfully
 - Verified that `Get-AppSourceProduct -authContext $ctx -productName '...' -silent` works
 - Verified that `Get-AppSourceProduct -authContext $ctx -productId '...' -silent` still works (no change)

- [x] Update ReleaseNotes.txt
- [x] Add telemetry
 - no additional telemetry needed


mmenninghaus:
@microsoft-github-policy-service agree [company="{SIEVERS-SNC Computer & Software GmbH & Co. KG}"]

mmenninghaus:
@microsoft-github-policy-service agree company="SIEVERS-SNC Computer & Software GmbH & Co. KG"

mmenninghaus:
Fix not sustainable.

PR #4163, opened 2026-06-17
Fix AS0054 when AppSourceCop runs on test apps

spetersenms:
### ❔What, Why & How

When enableCodeAnalyzersOnTestApps is set, Run-AlPipeline enables the AppSourceCop analyzer on test/bcpt apps but only generated AppSourceCop.json for regular apps, so test apps failed with AS0054 (no mandatory affixes configured).

The AppSourceCop.json creation now uses the same condition ($app -or $enableCodeAnalyzersOnTestApps) as the analyzer enablement, so the affixes config is written wherever the analyzer actually runs.

Related to issue: #

### ✅ Checklist

- [ ] Add tests (`Tests` / `LinuxTests`)
- [ ] Update ReleaseNotes.txt
- [ ] Add telemetry

### DIFF
--- a/AppHandling/Run-AlPipeline.ps1
+++ b/AppHandling/Run-AlPipeline.ps1
@@ -2348,7 +2348,7 @@ Write-Host -ForegroundColor Yellow @'
         }
     }
 
-    if ($enableAppSourceCop -and $app) {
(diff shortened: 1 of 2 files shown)


mazhelez (dismissed):
Release notes are missing

aholstrup1 (approved):

mazhelez (approved):

PR #4195, opened 2026-08-31
Wait for Docker service readiness before failing / running tests

mazhelez:
## Problem
CI test jobs intermittently fail early with:

> RuntimeException: Docker service is not yet ready.

Example: https://github.com/microsoft/navcontainerhelper/actions/runs/33378268075 (jobs `GetAndRunTests (version 17)`, `AppHandling`, `GetAndRunTests (latest)` all failed within ~1-3 min during "Run Tests").

## Root cause
On the Windows CI runners the Docker daemon can still be starting up when a test begins. `New-BcContainer` probes the daemon once with `docker version -f "{{.Server.Os}}/..."`; while the daemon is starting, `Server.Os` comes back empty and the function throws immediately instead of waiting.

## Fix
- **`ContainerHandling/New-NavContainer.ps1`** — poll the `docker version` probe for up to 120 seconds (wrapped in `try/catch` so a throwing docker client is also treated as "not ready") before giving up. Existing behavior is preserved: it still throws if the daemon never becomes ready, and still rejects non-Windows container mode.
- **`.github/workflows/RunTests.yaml`** — add a `Wait for Docker` step before `Run Tests` in the PS5 and PS7 (Windows) jobs so the daemon is ready before any test runs. The Linux job is unchanged.
- **`ReleaseNotes.txt`** — note added under 6.1.17.

This revives and rebases the earlier `copilot/analyze-docker-service-ready-issue` work onto current `main`, dropping its unintended `actions/checkout` downgrade and line-ending churn.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


spetersenms (dismissed):

ReleaseNotes.txt
```diff
@@ -1,6 +1,7 @@
 6.1.17
 Fix AppSource upload failing on Front Door SAS URIs
 BC29 moved dll to Admin folder
+New-BcContainer: Wait for the Docker service to become ready (poll up to 120 seconds) instead of failing immediately with "Docker service is not yet ready" when the daemon is still starting up (e.g. on CI runners). The RunTests workflow also waits for Docker before running the Windows test jobs.
```
aholstrup1: Releasenote is outdated now

aholstrup1 (dismissed):

aholstrup1 (approved):

jonasblunck (approved):

PR #4211, opened 2026-09-18
Fail test run when a test codeunit fails runtime compilation during test selection

mazhelez:
### ❔What, Why & How

**What.** Make `Run-TestsInBcContainer` fail the test run when a test codeunit fails **runtime (C#) compilation while the test runner selects tests**.

**Why.** When the test runner selects tests for an extension, the platform generates and compiles C# for every test codeunit in that extension. If a codeunit produces C# that does not compile (for example an AL construct the AL compiler accepts but the runtime C# generator rejects), the platform:

- logs the failure,
- **skips that codeunit** – none of its tests are discovered or run, and
- still completes the test run and reports **success**.

The only trace left behind is a file in the service instance's `apps\compilationerrors` folder. As a result an entire test codeunit can silently never run while the build stays green. We hit exactly this: a whole test codeunit was skipped and the pipeline passed.

**How.** `Run-TestsInBcContainer` now:

1. Snapshots the runtime compilation-error artifacts in the service instance's `...\MicrosoftDynamicsNavServer$<instance>\apps\compilationerrors` folder **before** running tests (so pre-existing artifacts, e.g. from publishing or a previous test app in a reused container, are ignored).
2. After the run, reads the folder again and reports any **new** artifacts (parsing the object declaration for a friendly name), emitting `::error::` / `##vso[task.logissue type=error]` when GitHubActions/AzureDevOps output is enabled.
3. **Throws** if any new artifact is found. This is deliberate: the tests never ran, so it must not be tolerable as a normal test failure (a plain `$allPassed = $false` would be downgraded to a warning under `TreatTestFailuresAsWarnings`). It mirrors the existing `"Validation errors occured"` throw.

The check is best-effort and only runs when a container is available (`$containerName`): online / `compilerFolder` runs are unaffected, and if the folder cannot be read it warns and continues so it never breaks unrelated builds.

Related to issue: #

### ✅ Checklist

- [ ] Add tests (`Tests` / `LinuxTests`)
- [x] Update ReleaseNotes.txt
- [ ] Add telemetry

> [!NOTE]
> Opened as a **draft**. The detection relies on the platform writing skipped test codeunits to the `apps\compilationerrors` folder; this was validated against a real build log but not yet in a live container run here. Feedback on the detection signal and placement is very welcome.


mazhelez:
@copilot resolve the merge conflicts in this pull request

mazhelez:
Closing this draft. After further investigation we confirmed the BCContainerHelper/NAV test client does not swallow hard runtime-compilation faults — it rethrows them (the NAV gate fails correctly). The remaining gap (individual `[Test]` methods silently dropped while their codeunit still reports success) is better handled in the BCApps build itself, as a declared-vs-executed reconciliation gate in `build/scripts/RunTestsInBcContainer.ps1`, alongside the existing test-tolerance and cancellation-detection safety nets. No change to navcontainerhelper is needed.

